PR #2, opened 2023-02-14
`leecher` - Fetch Events from Shotgrid

Minkiu:
The `leecher` is a service that will periodically, 10 seconds by default, "leech" events from Shotgrid and dispatch events in the Ayon event system of the topic `shotgrid.leech` where the whole of the Shotgrid payload is sent.


services/leecher/leecher/__main__.py
```diff
@@ -0,0 +1,9 @@
+import sys
+
+from .listener import ShotgridListener
```
iLLiCiTiT, line 3: I think `__main__.py` does not support relative imports (not sure).

services/leecher/leecher/listener.py
```diff
@@ -0,0 +1,234 @@
...
+        Args:
+            func (Callable, None): In case we want to override the default
+                function we cast to the processed events.
+        """
+        logging.info("Initializing the Shotgrid Listener.")
+
+        if func is not None:
+            self.func = func
+        else:
+            self.func = self.send_shotgrid_event_to_ayon
```
iLLiCiTiT, line 56:
```suggestion
        if func is None:
            func = self.send_shotgrid_event_to_ayon
        self.func = func
```

services/leecher/leecher/__main__.py
Minkiu: So far the code hasn't complained :D

PR #4, opened 2023-02-15
`processor` - Enroll `shotgrid.leech` events

Minkiu:
The processor will be enrolled to listen for `shotgrid.leech` events and start any function fouind in the `processor/hanbdlers` directory.

Each handler can register to a type of event, so that it only reacts to the enties that matters to it.


Minkiu:
I'm merging this but feel free to comment and I'll update it in the subsequent iterations, since I'm now working in the `processor` lib / handlers.

PR #5, opened 2023-02-16
`create_project` and `sync_from_shotgrid` handlers

Minkiu:
This commit implements two handlers, which are interconected.

Firstly there's the `create_project`, triggered when a `shotgrid.event`
with the `action` payload of `create-project` is dispatched, which
creates a Shotgrid project in Ayon, it will also replicate the Shotgrid
entities and tasks in Ayon, by creating folders and tasks that match
these in the project anatomy.
It also sets up the Shotgrid project creating custom fields (all
prepended by `sg_ayon_`) in Shotgrid and trigger a full Sync once
everything is in place.

Secondly, there's the `sync_from_shotgrid`, triggered when a
`shotgrid.event` with the `action` payload of `sync-from-shotgrid`
is dispatched, which will perform a full import of a Shotgrid project
into Ayon, it will fail if there are missing attributes in Shotgrid,
which should be present unless tampered manually, it won't go through
otherwise.

In both instances the actual syncronization is performed by the class
`SyncFromShotgrid` found in the `lib/sync_from_shotgrid.py` module, this
class performs the checks, and traverses the Shotgrid project in the look
for entities to be created or updated in Ayon.

A set of `constants` and `utils` are defined in the `lib` directory,
the former contains Shotgrid field codes, Ayon attributes names and
 mappings between the two, and the latter includes helper functions to
query and marshall data from Shotgrid, as well as creating entities in
Ayon.


services/processor/processor/handlers/create_project.py
```diff
@@ -0,0 +1,125 @@
...
+        return
+    else:
+        try:
+            logging.info(f"Creating project {project_name} in Ayon.")
+            ayon_project = create_project(project_name, project_code)
+        except Exception as e:
+            logging.error("Unable to create the Ayon project.")
+            log_traceback(e)
+            # We would normally retunr here, but thers a bug currently in `ayon_api`
+            # return
```
iLLiCiTiT, line 77:
```suggestion

    try:
        logging.info(f"Creating project {project_name} in Ayon.")
        ayon_project = create_project(project_name, project_code)
    except Exception as e:
        logging.error("Unable to create the Ayon project.")
        log_traceback(e)
        # We would normally retunr here, but thers a bug currently in `ayon_api`
        # return
```

services/processor/processor/handlers/sync_from_shotgrid.py
```diff
@@ -0,0 +1,67 @@
...
+    get_sg_project_by_name,
+)
+from processor.lib.constants import CUST_FIELD_CODE_SYNC
+
+
+from ayon_api import (
+    get_project,
+    get_project_anatomy_preset,
+    post
+)
```
iLLiCiTiT, line 16:
```suggestion
from ayon_api import get_project
```

services/processor/processor/handlers/create_project.py
```diff
@@ -0,0 +1,125 @@
+"""
```
iLLiCiTiT, line 1: I would rename the file to `project_sync.py` as `create_project` is more "actinable" name.

services/processor/processor/lib/sync_from_shotgrid.py
```diff
@@ -0,0 +1,282 @@
...
+            SHOTGRID_ID_ATTRIB
+        ).value
+
+        if not ayon_project_shotgrid_id_attrib:
+            self.log.error((
+                "Project creation probably went wrong, there's no shotgridID"
+                f"{ayon_project_shotgrid_id_attrib}"
+            ))
+            return
+        elif int(self._sg_project["id"]) != int(ayon_project_shotgrid_id_attrib):
```
iLLiCiTiT, line 160:
```suggestion
        if int(self._sg_project["id"]) != int(ayon_project_shotgrid_id_attrib):
```

services/processor/processor/lib/utils.py
```diff
@@ -0,0 +1,570 @@
...
+    ]
+    new_tasks = sg_tasks + project_entity.task_types
+    new_tasks = list({
+        task['name']: task
+        for task in new_tasks
+    }.values())
+
+    sg_entities = [
+        {"name": entity[0]}
+        for entity in get_sg_project_enabled_entities(
```
iLLiCiTiT, line 131:
```suggestion
        {"name": entity_type}
        for entity_type, _ in get_sg_project_enabled_entities(
```

services/processor/processor/lib/utils.py
```diff
@@ -0,0 +1,570 @@
...
+    sg_project_schema = sg_session.schema_entity_read(
+        project_entity=sg_project
+    )
+
+    project_navigation = sg_project["tracking_settings"]["navchains"]
+    project_navigation["Task"] = "entity"
+
+    project_entities = []
+
+    for entity in AYON_SHOTGRID_ENTITY_MAP:
```
iLLiCiTiT, line 498: I would recommend to change the variable to `entity_type`.

services/processor/processor/lib/utils.py
```diff
@@ -0,0 +1,570 @@
...
+) -> dict:
+    """ Get all Statuses on a Shotgrid project.
+
+    Args:
+        sg_session (shotgun_api3.Shotgun): Shotgun Session object.
+        shotgrid_project (dict): The project owning the Tasks.
+
+    Returns:
+        sg_statuses (dict): Shotgrid Project Statuses list.
+     """
```
iLLiCiTiT, line 532:
```suggestion
    """
```

services/processor/processor/lib/sync_from_shotgrid.py
Minkiu: I need this to run only if `ayon_project_shotgrid_id_attrib` is not empty, otherwise the `int` cast would fail.

iLLiCiTiT: But that handles the previous condition, which ends with `return`. So the `elif` is not needed because, if previous condition happens, it will never get to this line.

services/processor/processor/lib/utils.py
Minkiu: I agree, the problem is that SG entities are just dicts... I guess I could add `_dict` everywhere

iLLiCiTiT: I would not complain if it would be a dictionary, but this is string which holds entity type.

Minkiu: Aha, yeah, I got that, pushing a change to clarify this

fabiaserra (commented):
smart of me that I never submitted my comments, sorry @Minkiu

services/processor/processor/handlers/project_sync.py
```diff
@@ -0,0 +1,125 @@
...
+        logging.info(f"Creating project {project_name} in Ayon.")
+        ayon_project = create_project(project_name, project_code)
+    except Exception as e:
+        logging.error("Unable to create the Ayon project.")
+        log_traceback(e)
+        # We would normally retunr here, but thers a bug currently in `ayon_api`
+        # return
+
+    try:
+        ayon_project = get_project(project_name)
```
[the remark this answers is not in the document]
iLLiCiTiT: I think there was a bug in ayon api which would cause crash.

Minkiu:
No worries @fabiaserra , it's still evolving :D

PR #6, opened 2023-03-28
Implement Frontend and Endpoints.

Minkiu:
This commit implements the Addon frontend which can be accessed in the Ayon Server > Settings > Shotgrid Sync page.

The frontend offers two actions, import and sync.
For Importing, it will, via the Shotgrid REST API implemented in the callback for the `get-importable-projects` endpoint, fetch the projects that are not a template, and will allow the person to choose from a drodown the project to import, by clicking the "Import Project" button.

For Syncronizing, a call to the Ayon server GraphQl API is performed, and we only display the projects that have the `shotgridId` attribute filled, once the person picks a show and click the button "Sync Shotgrid Project".

In both instances, upong clicking the button, will dispacth an event of the topic `shotgrid.event` which has to be handled by a `processor` respectively, which are not included in this commit, these are catched by the `create-project` and `sync-from-shotgrid` endpoints.

The CSS and Javascript have been split in separate files.

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -1,12 +1,19 @@
+import json
+import socket
 from typing import Any, Type
 
 from ayon_server.addons import BaseServerAddon
+from ayon_server.api.dependencies import dep_current_user, dep_project_name
+from ayon_server.entities import UserEntity
+from ayon_server.events import dispatch_event
 from ayon_server.lib.postgres import Postgres
-
-from nxtools import logging
-
 from .settings import ShotgridSettings, DEFAULT_VALUES
 from .version import __version__
+
+from fastapi import Depends
+import requests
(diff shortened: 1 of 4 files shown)


server/__init__.py
```diff
@@ -1,25 +1,198 @@
...
+            logging.info(f"Dispatched event {event_id}")
+
+    async def _get_importable_projects(self):
+        """ Query Shotgrid for existing non-template Projects.
+
+        It uses the Rest APi to avoid importing the Shotgrid API.
+        """
+        addon_settings = await self.get_studio_settings()
+
+        if not addon_settings:
```
iLLiCiTiT, line 140: Maybe also check if shotgrid server is filled?

PR #8, opened 2023-04-03
`update_from_shotgrid` handler on Shotgrid Events

Minkiu:
This commit implements the `update_from_shotgrid` handler, triggered when a `shotgrid.event` with the `action` payload of `shotgrid-event` is dispatched, which will attempt to either `create`, `update` or `remove` an Ayon entity, that corresponds to the Shotgrid entity that the event is acting on.

Shotgrid has the ability to "retire" (move to trash) entities, and then bring them back, in those instances we treat it as if we are creating a new entity, since we don't have that feature in Ayon.

As of this implementation, the `update_entity` only cares for the entity name since it's all we really track at the moment, and ensures the `id` mapping between Ayon and Shotgrid is still valid.


services/processor/processor/lib/update_from_shotgrid.py
```diff
@@ -0,0 +1,277 @@
...
+        )
+
+        ay_entity = None
+
+        if not sg_entity_dict.get(CUST_FIELD_CODE_ID):
+            self.log.warning("Shotgrid Missing Ayon ID")
+
+            # Attempt to find it by name
+            if sg_entity_dict["type"].lower("task"):
+                ay_entity = self._entity_hub.get_entity_by_name(
```
iLLiCiTiT, line 200: I don't think there is `get_entity_by_name` method?

services/processor/processor/lib/update_from_shotgrid.py
```diff
@@ -0,0 +1,277 @@
...
+    def remove_entity(self):
+        """Try to remove an entity in Ayon.
+        """
+        sg_entity_dict = get_sg_entity_as_ay_dict(
+            self._sg_session,
+            self._sg_event["entity_type"],
+            self._sg_event["entity_id"],
+        )
+
+        ay_entity = None
```
iLLiCiTiT, line 249: You should not allow to remove entity which has published subsets. You should check `ay_entity.immutable_for_hierarchy` before removement and handle that in some way.

PR #11, opened 2023-04-14
Implement the client side of the Addon

Minkiu:

PR #12, opened 2023-04-25
Client small tweaks

iLLiCiTiT:
## Description
Small tweaks in collect session and entities in publish plugins. Added missing variable in collecte session and changed label. Use `KnownPublishError` instead of `ValueErrors`. Replace usage of lists with sets where possible to avoid collecting the same information more times. Shotgrid tasks are collected based on parent id and task names instead of looking to ayon.

Moved import of `credentials` to method where is used.

Modifications for PR https://github.com/ynput/ayon-shotgrid/pull/11

### DIFF
--- a/client/ayon_shotgrid/addon.py
+++ b/client/ayon_shotgrid/addon.py
@@ -1,5 +1,7 @@
 import os
 
+from .lib import credentials
+
 from openpype.modules import (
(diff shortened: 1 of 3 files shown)


iLLiCiTiT:
Already in https://github.com/ynput/ayon-shotgrid/pull/11

PR #13, opened 2023-05-10
fix sg project attrs constant

Iskander7d:
While syncing SG project we faced the issue with ayon project name. Ayon tried to get field "sg_ayon_project_name" from SG Project that was not set before by "create_ay_fields_in_sg_project" method

### DIFF
--- a/services/processor/processor/lib/constants.py
+++ b/services/processor/processor/lib/constants.py
@@ -29,6 +29,11 @@ SG_PROJECT_ATTRS = {
         "name": "Ayon Server URL",
         "type": "text",
         "sg_field": CUST_FIELD_CODE_URL,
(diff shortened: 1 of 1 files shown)


Minkiu:
Hey there! 
Sorry for having missed this!

I have now merged a PR https://github.com/ynput/ayon-shotgrid/pull/19 which does away with this field, since in AYON the project id is the actual project name! So when synchronizing, it will now try to create a `sg_ayon_id` in the Project schema. More specifically in this commit: https://github.com/ynput/ayon-shotgrid/commit/ac18ad2996ea45868c76d2031c151ef9ad275a61

Sorry for the inconvenience, and feel free to add me as reviewer next time :)

PR #14, opened 2023-05-26
New `settings.shotgrid_project_code_field` field

Minkiu:
Since it seems not all SG instances have the `code` field for projects, a new setting on the addon (which defaults to `code`) allows the person to define which field in shotgrid to use as `code` to create projects in Ayon.


iLLiCiTiT:
In this case it is hard to decide if we want to force it. Can we create the attribute in shotgrid based on the settings?

Can't test it but looks valid and reasonable.

iLLiCiTiT (approved):

PR #16, opened 2023-06-21
Convert into a Rez package

Minkiu:
This commit adds the `package.py` and `rezbuild.py` files necessary to allow rez to build and release it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,7 +4,7 @@ __pycache__/
 *.py[cod]
 *$py.class
 .env
-
+build/*
 package/*
 .env/*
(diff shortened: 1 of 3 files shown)


kalisp:
How would I run this through Rez (and maybe pass `--output_dir` argument)?

Minkiu:
I'll write a proper tutorial, but you do not need to specify the output directory, and rez should deploy either locally or to the actual configured rez repository.

jakubjezek001:
We have decided not to use REZ for now. Closing it for now and will reopen it in future if we will decide to continue with the design.

PR #24, opened 2023-08-25
`client` - Fix wrong method when collection SG session

Minkiu:
The `collect_shotgrid_session` was using a wrong method to create the SG session, this commit ensures it uses the correct one `create_shotgrid_session`.

Also removed the trailing dot from the `pyblish` plugins name so it follows all the other plugins.

### DIFF
--- a/client/ayon_shotgrid/plugins/publish/collect_shotgrid_session.py
+++ b/client/ayon_shotgrid/plugins/publish/collect_shotgrid_session.py
@@ -8,7 +8,7 @@ class CollectShotgridSession(pyblish.api.ContextPlugin):
     """Collect shotgrid session using user credentials"""
 
     order = pyblish.api.CollectorOrder
-    label = "Collecting Shotgrid session."
+    label = "Collecting Shotgrid session"
 
     def process(self, context):
         user_login = os.getenv("AYON_SG_USERNAME")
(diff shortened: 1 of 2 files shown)


client/ayon_shotgrid/plugins/publish/collect_shotgrid_session.py
```diff
@@ -17,23 +17,25 @@ def process(self, context):
...
-
         try:
-            sg_session = shotgrid_module.get_shotgrid_session(user_login)
+            sg_session = shotgrid_module.create_shotgrid_session(user_login)
             self.log.info("Succesfully logged in into the Shotgrid API.")
-        except Exception:
+        except Exception as e:
             raise KnownPublishError(
-                "Could not connect to Shotgrid {} with user {}".format(
+                "Could not connect to Shotgrid {0} with user {1} with error:\n{2}".format(
```
iLLiCiTiT, line 35: This is message for user. Not sure how much will the error message help him? Maybe rather use log for more details and keep the message simple for user:
```
except Exception:
    self.log.error("Failed to connect to shotgrid.", exc_info=True)
    raise KnownPublishError((
            "Could not connect to Shotgrid {} with user {}"
    ).format(shotgrid_url, user_login))
```

This is based on premise that shotgrid exceptions are not "artist understandable".

Minkiu: Yeah, sound better.

PR #25, opened 2023-09-14
Implement `AyonShotgridHub`

Minkiu:
The idea is to make the interaction between Shotgrid and Ayon easier, so a class `AyonShotgridHub` that will handle common actions has been implemented, and any parts that needed to interact with it have been updated, the actual change log is below:

* `services` - Bump `ayon-python-api` and `shotgun-api3` versions
* `processor` Attempt to slim down the resulting image Remove unnecessary pacakges and cache to try slim down the resulting image.
* `frontend` - Fix JS `syncShotgridToAyon` not including `project_code`
*  `leecher` - Add `proejct_code` to the event Since we are moving towards using the `AyonShotgridHub` and it requires both the `project_name` and `project_code` the `leecher` has been modified to include the `project_code` in the disaptched event payload.
* `ShotgridProcessor` - Remove SG initialization and improve docstring Moving forward, we'll be using the `AyonShotgridHub` to check both in AYON and Shotgrid, so the logic where we check the Shotgun connection has been defered to `AyonShotgridHub`.
* `processor.AyonShotgridHub` a class to interact with Ayon and Shotgrid Influenced by the `ayon_api.EntityHub` the `AyonShotgridHub` is a class that provided a valid Project name and code, will perform all the necessary checks and provide methods to keep an Ayon and Shotgrid project in sync.
* `processor` - Shuffle logic from `lib` to `ayon_shotgrid_hub` Most of the `lib` logic has now been moved as part of the `AyonShotgridHub`, so both `constants` and `utils` have been promoted on level up (in another commit), this commit deletes the no longer needed files.
* `processor` - Implement use of `AyonShotgridHub` in `processor.handlers` Removed outdated handlers and confomed all the logic in two, which make use of the `AyonShotgridHub` simplifying the overall logic.


services/processor/processor/ayon_shotgrid_hub/update_from_shotgrid.py
```diff
@@ -0,0 +1,221 @@
...
+    sg_entity_dict = get_sg_entity_as_ay_dict(
+        sg_session,
+        sg_event["entity_type"],
+        sg_event["entity_id"],
+    )
+
+    if not sg_entity_dict.get(CUST_FIELD_CODE_ID):
+        logging.warning("Shotgrid Missing Ayon ID")
+
+    ayon_entity_hub.query_entities_from_server()
```
iLLiCiTiT, line 149: You don't have to query whole project (that may take some time). You can use `get_folder_by_id`, `get_task_by_id` or `get_or_query_entity_by_id`.

services/processor/processor/ayon_shotgrid_hub/update_from_shotgrid.py
```diff
@@ -0,0 +1,221 @@
...
+
+    if not ay_entity.immutable_for_hierarchy:
+        ayon_entity_hub.delete_entity(ay_entity)
+    else:
+        logging.info("Entity is immutable.")
+        ay_entity.attribs.set(SHOTGRID_ID_ATTRIB, SHOTGRID_REMOVED_VALUE)
+
+    ayon_entity_hub.commit_changes()
+
+
```
iLLiCiTiT, line 221:
```suggestion

```

services/processor/processor/processor.py
```diff
@@ -7,81 +7,77 @@
...
 
-        # Grab all the `handlers` from `processor/handlers` and map them to
-        # the events that are meant to trigger them
         self.handlers_map = None
 
         try:
             self.settings = ayon_api.get_addon_settings(
                 os.environ["AYON_ADDON_NAME"],
                 os.environ["AYON_ADDON_VERSION"]
             )
```
iLLiCiTiT, line 44:
```suggestion
            self.settings = ayon_api.get_service_addon_settings()
```

services/processor/processor/ayon_shotgrid_hub/update_from_shotgrid.py
Minkiu: I was doing this since I didnt know if it was a task or a folder just yet, `get_or_query_entity_by_id` should do the trick tho! thanks

PR #30, opened 2023-09-27
OP-6918 - Shotgrid 0.2.1 - Papercuts fixes

Minkiu:
**WARNING**: I do `git rebase` and `git push --force-with-lease` if you clone this branch, make sure you pull with `git pull --rebase`.

Here are the most annoying papercuts that made working with Shotgrid a bit difficult:

* `server` - Use the `version.py` to define services Previously the version of the `service` docker image was hardcoded, this commit reads whatever is set in the `version.py` of the addon.

* Use AYON `secrets` for the Shotgrid API credentials We were storing the credentials for the Shotgrid API as plaintext, this change now makes use of the `secrets` backend in AYON.

* `leecher` Rename variables to use `sg_` Before the commit we were prefixing variables with `shotgrid_`, we now use `sg_` across all codebases, so this commit aligns the `leecher` with the rest.

* `frontend` Correctly merge AYON and SG projects We were blindly merging the two arrays of projects, with this commit we check if the project already exists and if so we update it, otherwise we add it to the array of `allProjects`.

* `services.utils` Read Pipeline Steps for Tasks Instead of querying for all the tasks in a project, we now read the project schema to get all the "Pipeline Steps" which is a more realistic representations of "Task".

* `services.ayon_shotgrid_hub` Commit changes after each new entity in sync When performing a sync with "shotgrid" as source, we want to commit after each entity is created so if the `processor` fails at any point we don't loose any potential progress.

* `services.leecher` - Leech only projects that have AYON auto sync Previously the user was required to provide a comma separated list of project codes so that the leecher would not get events from other projects which we don't want to sync and that potentially do not exist in ayon, so now the leecher filters all the projects be the `sg_ayon_auto_sync` field.

* `frontend` - Vendor `axios` and fix links We are now vendoring `axios` instead of loading from a third party CDN.

*  `services` - `Makefile` now sets `ADDON_` env vars Instead of defininf the ENV VARS in the `.env` file, we now set them directly from the `Makefile`, so they are always in sync with the `version.py` file.

* `services.processor` - Update event description A better description of the event so that the user can get some kind of (limited) feedback on what the `processor` is doing, this should improve once `ash` has better capabilities of surfacing the spwaned services logs.

* Update `README`

* `client` Rework Shotgrid login Since we now use the secrets in the server side, we can no longer get the api key to login, so we need to ask que user for their user and password; this commit makes the changes so we store the login as a tuple of (`username`, `password`) in the `OpenPypeSettingsRegistry`, and use that entry to login with Shotgrid.


iLLiCiTiT (approved):
Looked and the code, and the changes make sense. Should be tested thou....

services/processor/processor/processor.py
```diff
@@ -155,11 +166,11 @@ def start_processing(self):
                     except Exception as e:
                         logging.error(f"Unable to process handler {handler.__name__}")
                         log_traceback(e)
-                        ayon_api.update_event(event["id"], status="finished")
-                        ayon_api.update_event(source_event["id"], status="finished")
+                        ayon_api.update_event(event["id"], status="failed", description=f"An error ocurred while processing the Event: {e}", summary="Feiled procesing the event Handler...")
+                        ayon_api.update_event(source_event["id"], status="failed", summary=f"The service `processor` was unable to process this event. Check the `shotgrid.proc` <{event['id']}> event for more info.")
```
iLLiCiTiT, line 170:
```suggestion
                        ayon_api.update_event(
                            event["id"],
                            status="failed",
                            description=f"An error ocurred while processing the Event: {e}",
                            summary="Feiled procesing the event Handler...")
                        ayon_api.update_event(
                            source_event["id"],
                            status="failed",
                            summary=f"The service `processor` was unable to process this event. Check the `shotgrid.proc` <{event['id']}> event for more info.")
```

services/processor/processor/processor.py
```diff
@@ -155,11 +166,11 @@ def start_processing(self):
...
 
                 logging.info("Event has been processed... setting to finished!")
-                ayon_api.update_event(event["id"], status="finished")
+                ayon_api.update_event(event["id"], description="Event processed succsefully.", status="finished")
```
iLLiCiTiT, line 173:
```suggestion
                ayon_api.update_event(
                    event["id"],
                    description="Event processed succsefully.",
                    status="finished")
```

server/__init__.py
```diff
@@ -29,9 +29,9 @@ class ShotgridAddon(BaseServerAddon):
     frontend_scopes: dict[str, Any] = {"settings": {}}
 
     services = {
-        "ShotgridLeecher": {"image": "ynput/ayon-shotgrid-leecher:0.2.0"},
-        "ShotgridProcessor": {"image": "ynput/ayon-shotgrid-processor:0.2.0"},
-        "ShotgridTransmitter": {"image": "ynput/ayon-shotgrid-transmitter:0.2.0"},
+        "ShotgridLeecher": { "image": f"ynput/ayon-shotgrid-leecher:{__version__}"},
+        "ShotgridProcessor": { "image": f"ynput/ayon-shotgrid-processor:{__version__}"},
+        "ShotgridTransmitter": { "image": f"ynput/ayon-shotgrid-transmitter:{__version__}"},
```
iLLiCiTiT, line 34:
```suggestion
        "ShotgridLeecher": {"image": f"ynput/ayon-shotgrid-leecher:{__version__}"},
        "ShotgridProcessor": {"image": f"ynput/ayon-shotgrid-processor:{__version__}"},
        "ShotgridTransmitter": {"image": f"ynput/ayon-shotgrid-transmitter:{__version__}"},
```

PR #31, opened 2023-10-10
Package creation: Fix pyproject.toml file location

iLLiCiTiT:
## Description
Copy `pyproject.toml` file next to zipped client code instead of adding it inside the zip.

### DIFF
--- a/create_package.py
+++ b/create_package.py
@@ -186,9 +186,7 @@ def zip_client_side(addon_package_dir, current_dir, log):
     """
 
     client_dir = os.path.join(current_dir, "client")
-    if not os.path.isdir(client_dir):
-        log.info("Client directory was not found. Skipping")
-        return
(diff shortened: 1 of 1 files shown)


create_package.py
```diff
@@ -186,9 +186,7 @@ def zip_client_side(addon_package_dir, current_dir, log):
     """
 
     client_dir = os.path.join(current_dir, "client")
-    if not os.path.isdir(client_dir):
-        log.info("Client directory was not found. Skipping")
-        return
+    addon_subdir_path = os.path.join(client_dir, ADDON_CLIENT_DIR)
```
Minkiu, line 189: I assume we check somewhere else that client dir exists

iLLiCiTiT: Nope, that was "universal" code for base addons, in shotgrid case we 100% know there should be client dir...

PR #39, opened 2023-10-20
Fix versions not in correct folder and double root for PublishedFiles

Minkiu:
*  `client` Bump Shotgrid API version to v3.4.0
*  `ayon_shotgrid_hub` Better logging when creating from SG event
* `services.utils` Version's parent field is always entity
* `client` Fix duplicated root for local files On Windows, when creating a `PublishedFile` the server was prepending the project root to the filepath which it already had it, this commit will read the first Local Storage defined in Shotgrid and attempt to get a partial path of the file, which is then used to create the `PublishedFile` and avoids the double root.
*  `service` Update the Shotgrid API to v3.4.0
* `Makefile` fix removing existing docker image The previous command to find container images wasn't really working, with this change we now ensure we are deleting the local image if there's any found with the same tag.
* `services` Pin base image to `python:3.11-alpine3.18` Python Docker images maintainers decided to upgrade to Python 3.12 for the Alpine 3.18 images, even though the default Python in Alpine is 3.11.
* `services` Bump `ayon-python-api` verion to `^0.5`
* `services` Fix variable typo in `__main__` I mistyped "shotgird" in the initial implementation, this commit fixes the three instances of the typo.
* `services` - Add `max_retries` to `enroll_event_job` The server won't allow to enroll with a null value for `max_retries` when enrolling jobs, we now specify it, and set it to two, this could potentially be added to the Shotgrid settings.
* `shotgrid_common` Fix Tasks creation form SG When we made the change to use Pipeline Steps for the tasks creation, we didn't change the `utils` to get the correct name of the Task (the step name), which then it made it fail when attempting to create taks, since we were still using the SG task name, which didn't match any task in AYON.
* `services` Improve logging for `transmitter` and `leecher`
* `ayon_shotgrid_hub` Improve Sync for "parentless" entities Fixed when we create entities in Shotgrid which parents should be the Project, instead of assigning the project as a Sequence for a Shot for example.

# WARNING: If you cloned this branch pull changes with `git pull --rebase`

### DIFF
--- a/client/ayon_shotgrid/plugins/publish/integrate_shotgrid_publish.py
+++ b/client/ayon_shotgrid/plugins/publish/integrate_shotgrid_publish.py
@@ -3,6 +3,7 @@ import platform
 
 import pyblish.api
 
+from openpype.pipeline import KnownPublishError
 from openpype.pipeline.publish import get_publish_repre_path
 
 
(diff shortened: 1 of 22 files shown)


Minkiu:
I'm gonna test that the three services build and work, and perform a publish before merging and making a new release.

Minkiu:
Found more issues, specially when creating Tasks types, which now its almost fixed, but the backend/frontedn seems to be doing something weird, so it breaks the `transmitter`

client/ayon_shotgrid/plugins/publish/integrate_shotgrid_publish.py
```diff
@@ -52,15 +53,51 @@ def process(self, instance):
                 query_filters
             )
 
-            if platform.system() == "Windows":
-                local_path = local_path.replace("\\", "/")
+            sg_local_store = sg.find_one(
```
antirotor, line 56: This throws `NameError`, it should be

```suggestion
            sg_local_store = sg_session.find_one(
```

antirotor (changes requested):
I've hit this:

```
shotgun_api3.shotgun.Fault: API read() invalid/missing string entity 'type':
{"type"=>"LocalStore",
 "return_fields"=>["mac_path", "windows_path", "linux_path"],
 "filters"=>{"logical_operator"=>"and", "conditions"=>[]},
 "return_only"=>"active",
 "paging"=>{"entities_per_page"=>1, "current_page"=>1},
 "api_return_image_urls"=>true,
 "return_paging_info_without_counts"=>false}
 ```
in `integrate_shotgrid_publish.py` - I guess that it's because I don't have `localStore` set on the SG project. We should handle this case somehow.

Minkiu:
Okay, I fixed the bug spotted by @antirotor and a plethora of errors that surfaced, since the PR is becoming unwieldy, I'll merge and make a release tomorrow first thing, so we can rebase other PRs and @m-u-r-p-h-y can test stuff on his end too.

Minkiu:
> I've hit this:
> 
> ```
> shotgun_api3.shotgun.Fault: API read() invalid/missing string entity 'type':
> {"type"=>"LocalStore",
>  "return_fields"=>["mac_path", "windows_path", "linux_path"],
>  "filters"=>{"logical_operator"=>"and", "conditions"=>[]},
>  "return_only"=>"active",
>  "paging"=>{"entities_per_page"=>1, "current_page"=>1},
>  "api_return_image_urls"=>true,
>  "return_paging_info_without_counts"=>false}
> ```
> 
> in `integrate_shotgrid_publish.py` - I guess that it's because I don't have `localStore` set on the SG project. We should handle this case somehow.

That's something that needs to be set up at a Shotgrid instance level, so I would assume by the time someone tries to publish they already have it enabled? Are you publishing against `https://ynput.shotgrid.autodesk.com` ?

Minkiu:
Just tested the publish and it worked:
image

Issue #22, opened 2023-08-17
Unable to Import Project - Validation Error


matteoveglia:
EDIT:
For anyone who finds this later, it contains some further info on how the ShotGrid service connects to ShotGrid and AYON, what data is synchronised, limitations and how to best launch the services required.

-------------

Hi,

I've finally managed to add the ShotGrid addon and get the services to function correctly.
I followed all the steps in the readme, build the docker images but I've run them detached (edited the MakeFile).

The addon fetches from our ShotGrid just fine, it lists the projects correctly but when I go to hit import I get this error:
![image](https://github.com/ynput/ayon-shotgrid/assets/1258632/0d1a6f3e-07df-4e5c-8e31-bb1f43a2d7b1)

Then, checking the event log, this is the string of events that happen:
![image](https://github.com/ynput/ayon-shotgrid/assets/1258632/fa2d14ec-6e03-4636-af2f-910badc4bea3)

Here's the output of what AYON sees from the project listing (pre-import):
![image](https://github.com/ynput/ayon-shotgrid/assets/1258632/ca4cf642-69b0-4f5e-bc75-3b81b14b3b84)

Here's the output of the validation error:
```shell
{
  "id": "1cc7f2b83cad11ee84a40242ac120005",
  "hash": "1cc7f2b83cad11ee84a40242ac120005",
  "topic": "log.error",
  "sender": null,
  "project": null,
  "user": null,
  "dependsOn": null,
  "status": "finished",
  "retries": 0,
  "description": "Validation error",
  "summary": {},
  "payload": {
    "message": "Validation error\n1 validation error for Request\npath -> project_name\n  string does not match regex \"^[a-zA-Z0-9_][a-zA-Z0-9_\\.\\-]*[a-zA-Z0-9_]$\" (type=value_error.str.regex; pattern=^[a-zA-Z0-9_][a-zA-Z0-9_\\.\\-]*[a-zA-Z0-9_]$)"
  },
  "createdAt": "2023-08-17T03:21:10.494161+00:00",
  "updatedAt": "2023-08-17T03:21:10.494161+00:00"
}
``` 

And here's the output from the ShotGrid Processor docker container
![image](https://github.com/ynput/ayon-shotgrid/assets/1258632/1da0e4c6-6736-4468-bfbc-f1c420d871ac)

Here's the thing, I get that the projectCode can't have any special characters, spaces etc but it doesn't... though it seems clear to me that the processor is trying to use the full projectName instead.

[...]

PR #50, opened 2023-11-03
Fixes for version 0.2.9

Minkiu:
Changes for next version:

* `leecher` Fix query not filtering projects correctly: The filter for the Events table wasn't correctly filtering out projects which did not have the "AYON Auto Sync" checkbox enabled.
* `settings` Use only one field for the secret: When specifying the secret containing the Shotgrid API key, we now only display one field, which is to select the secret, instead of two where you had to choose the same secre twice.
* `services` Honor `shotgrid_project_code_field` setting: With the implementation of the AyonShotgridHub I missed passing the setting to it so the setting wasn't being honored, this commit ensures we pass it in the `leecher` event and we take it in account in both the `processor` and the `transmitter`.
* `services` Fix `utils` not finding the parent entity on complex hierarchies: In Shotgrid you can set hierarchies of multiple levels, for example, the hierarchy of a Shot could be "Seq > Secene > Shot"; when doing this, the "navchains" gert represented such as 'sg_scene,Scene.sg_sequence'; so this commit ensures we understand these, and that we pick the first element, since thats the parent based on the hierarchy.


Minkiu:
I'm waiting on https://github.com/ynput/ayon-shotgrid/issues/22#issuecomment-1794818705 to see if I need to fix something on the Scene side, since I cant replicate it on my end, but will add some fields sync in a new commit.

Minkiu:
Allright, made the fix from the github issue, will merge and deploy a new version; and can tackle the status and attributes sync in another PR.

PR #58, opened 2023-12-01
Use SG API key for creating session and remove login functionality

fabiaserra:
This removes the need for logging in to SG through the Tray and just make use of the API key + `sudo_as_login` to integrate into SG. We are already doing extra validations of the user if it exists in SG to also need them to type yet another password... which in some studios might have the 2FA authentication and make this even more complicated. Unless we can enable SSO authentication across Ayon and SG I don't see any value for us on enabling this. On our use case I make use of `os.getenv("USER")` instead of `AYON_SG_USER` but on this case we might need them to be able to type what's their SG user if for any reason that differs from the Ayon username.

On the other hand, I bring here some of the changes that we have done on the OP shotgrid module to make it work better for our needs:

* Avoid creating SG versions/published files unless there's actually some data to update
* Populate some extra fields on the SG versions coming from Ayon
* Remove version from PublishedFiles so they are grouped together
* Change the name of the SG version code to be specific to the subset (although as noted on the code already this should be coming from the settings instead)
* Allow 'admin' roles to publish to any SG project


client/ayon_shotgrid/plugins/publish/integrate_shotgrid_publish.py
```diff
@@ -141,3 +168,43 @@ def process(self, instance):
                 )
             instance.data["shotgridPublishedFile"] = sg_published_file
 
+    def _find_published_file_type(self, instance, filepath, representation):
```
fabiaserra, line 171: this function is a pretty simplistic way to set some types on the PublishedFiles of SG and it relies on setting these on your SG instance:
Screenshot 2023-12-01 at 1 38 55 PM

matteoveglia: Yeah this is great too, how would you select what SH publish type (representation) you're publishing too?
Do I need to make sure I name the SG and AYON representations exactly the same?

client/ayon_shotgrid/addon.py
```diff
@@ -33,26 +40,13 @@ def get_plugin_paths(self):
     def create_shotgrid_session(self):
         from .lib import credentials
 
-        sg_username, sg_password = credentials.get_local_login()
-
-        if not sg_username or not sg_password:
-            return None
+        sg_username = os.getenv("AYON_SG_USERNAME")
```
Minkiu, line 43: Where do you normally define this environment variable, and what should happen if it's not set empty?

client/ayon_shotgrid/plugins/publish/validate_shotgrid_user.py
```diff
@@ -20,28 +20,29 @@ def process(self, context):
...
         sg_user = sg_session.find_one(
             "HumanUser",
             [
                 ["login", "is", user_login],
                 ["projects", "name_contains", project_name]
             ],
-            ["projects"]
+            ["projects", "permission_rule_set"]
         )
+        admin = sg_user["permission_rule_set"]["name"] == "Admin"
```
Minkiu, line 33: This probably should be something like:
`admin = True if sg_user["permission_rule_set"]["name"] == "Admin" else False`

But I believe that line 29 might prevent from an admin  to be shown

client/ayon_shotgrid/addon.py
fabiaserra: on my case I'm actually just using `$USER` as all our artists machines have that defined so we don't need any extra user input to define their login for SG but that's why I left it open to discuss. We could potentially expose an option on the settings where admins can choose any existing env var to use as SG login and/or optionally expose the login menu so each user can set it themselves on the registry

fabiaserra:
> Hey! Thanks for the contribution :D
> 
> I just got around trying this, and I'm getting an error when trying to publish:
> 
> Setting `AYON_SG_USERNAME=admin`, ended up in:

And are you sure you have an `admin` user in your SG server?

client/ayon_shotgrid/plugins/publish/validate_shotgrid_user.py
fabiaserra: Yeah that's more explicit and clear, I will update it to your snippet, thank you!

fabiaserra: done here https://github.com/ynput/ayon-shotgrid/pull/58/commits/9bc1af93b2c5713b961cd7a776924640cd80a06e

Minkiu:
Thanks for your contribution! Merging.

Minkiu (approved):

fabiaserra:
> Not sure why the login functionality was removed. This is not working and breaking the addon completely . . .

it works for us while the login functionality didn't. You just need to provide a way to fill up the user env var, either automatically through the Ayon login logic or on the general environments. Or if you'd rather bring that login dialog back, it should be just another optional way to provide the user login

m-u-r-p-h-y:
there is no mention of any user env in the Readme file. 

good to hear it is working for you, but everyone else needs to dig through the code to figure out how to log in to Shotgrid and why it is not working.

fabiaserra:
> there is no mention of any user env in the Readme file.
> 
> good to hear it is working for you, but everyone else needs to dig through the code to figure out how to log in to Shotgrid and why it is not working.

really the same counter argument with the prior iteration with the login -> no readme and not working haha I think we just need to put some more hands on this codebase and give it some love for the rest, as it is it's not in a state of plug and play and have fun, it's never been for SG <-> OP/Ayon

jakubjezek001:
I am considering the feasibility of reinstating the original functionality that was removed in this pull request, within the context of the addon studio settings. My recommendation would be to have both options available, but with the environment variable workflow being the default. However, if anyone wishes to utilize the logging functionality from AY Desktop tray menu, they could easily activate it.

m-u-r-p-h-y:
The question is what is the SG user good for actually, if all interaction is happening using sg_script API key. Why is it different from ftrack integration? How do we deal with different user permissions in Ayon? Does it mean a user with limited permissions in Ayon has unlimited access to SG? etc..

fabiaserra:
> The question is what is the SG user good for actually, if all interaction is happening using sg_script API key. Why is it different from ftrack integration? How do we deal with different user permissions in Ayon? Does it mean a user with limited permissions in Ayon has unlimited access to SG? etc..

How is the ftrack integration?

A user with limited permissions in Ayon won't really have unlimited access to SG as the frontend is blocking him from accessing any of the projects he doesn't have access. Of course a technical artist could always figure out the API to grab the session and do queries or publishes to any project with the API key but I wouldn't worry about that, it's really the same paradigm as in Ayon once you know about the API key

PR #62, opened 2023-12-15
Use SG pipeline step as task type and respect SG task name

fabiaserra:
There was an issue where SG Task names weren't being used and instead only the SG Pipeline Steps were populating Ayon. This PR changes that so the SG Pipeline Steps are only used to fill up the Project task types


Minkiu (approved):
All good, thanks for your contribution! Mergin it

PR #64, opened 2024-01-19
New release and bug fixes

Minkiu:
* Update services to user `ayon-python-api-1.0.0` And bump addon version to "0.3.0"
* Add new clean-build-all to Makefile
* `services.processor` reduce debug loggin noise
* `services.transmitter` Honor service polling rate and reduce logging noise
* `services.transmitter` Set event to failed if it errors
*  `ayon_shotgrid_hub` handle errors on entity creation
* `services.leecher` Reduce the logging noise

### DIFF
--- a/server/settings/main.py
+++ b/server/settings/main.py
@@ -2,6 +2,7 @@ from pydantic import Field
 
 from ayon_server.settings import BaseSettingsModel
 from ayon_server.settings.enum import secrets_enum
+from ayon_server.graphql.resolvers.users import get_users
 
 
 class ShotgridServiceSettings(BaseSettingsModel):
@@ -14,6 +15,12 @@ class ShotgridServiceSettings(BaseSettingsModel):
         default=10,
         title="How often (in seconds) to process Shotgrid related events.",
         validate_default=False,
+    )
+
+    ayon_service_user: str = Field(
+        default="service",
+        title="AYON service user",
+        description="The AYON user used in the services (the user corresponding to the `AYON_API_KEY` set in the service)",
     )
 
 
--- a/services/leecher/leecher/listener.py
+++ b/services/leecher/leecher/listener.py
@@ -99,7 +99,6 @@ class ShotgridListener:
         sg_projects = self.sg_session.find(
             "Project", filters=[["sg_ayon_auto_sync", "is", True]]
         )
-        logging.debug(f"Projects with the autosync enabled {sg_projects}")
 
         if not sg_projects:
             return []
@@ -165,7 +164,6 @@ class ShotgridListener:
         while True:
             sg_filters = self._build_shotgrid_filters()
             if not sg_filters:
-                logging.debug("No Shotgrid project with `Ayon Sync` enabled.")
                 time.sleep(self.shotgrid_polling_frequency)
                 continue
 
@@ -181,10 +179,6 @@ class ShotgridListener:
                 )
 
                 if not events:
-                    logging.info("No new events found.")
-                    logging.info(
-                        f"Waiting {self.shotgrid_polling_frequency} seconds..."
-                    )
                     time.sleep(self.shotgrid_polling_frequency)
                     continue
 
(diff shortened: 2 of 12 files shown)


services/leecher/leecher/listener.py
```diff
@@ -181,10 +179,6 @@ def start_listening(self):
                 )
 
                 if not events:
-                    logging.info("No new events found.")
```
fabiaserra, line 184: 👏 gracies

services/processor/processor/processor.py
```diff
@@ -118,10 +118,11 @@ def start_processing(self):
         will trigger the `handlers/project_sync.py` since that one has the attribute
         REGISTER_EVENT_TYPE = ["create-project"]
         """
-        logging.info("Start enrolling for Ayon `shotgrid.event` Events...")
```
fabiaserra, line 121: debug?

services/transmitter/transmitter/transmitter.py
```diff
@@ -67,6 +74,11 @@ def start_processing(self):
             "entity.folder.attrib_changed",
         ]
 
+        logging.info(
```
fabiaserra, line 77: debug?

services/transmitter/transmitter/transmitter.py
```diff
@@ -123,13 +131,19 @@ def start_processing(self):
                 ay_project = ayon_api.get_project(project_name)
 
                 if not ay_project:
-                    logging.error(f"Project {project_name} does not exit in AYON")
+                    # This should never happen since we only fetch events of 
+                    # projects we have shotgridPush enabled; but just in case
+                    # The event happens when after we deleted a project in AYON.
+                    logging.error(
+                        f"Project {project_name} does not exit in AYON "
```
fabiaserra, line 138:
```suggestion
                        f"Project {project_name} does not exist in AYON "
```

server/settings/main.py
```diff
@@ -16,6 +17,12 @@ class ShotgridServiceSettings(BaseSettingsModel):
         validate_default=False,
     )
 
+    ayon_service_user: str = Field(
+        default="service",
+        title="The AYON Shotgird user",
```
fabiaserra, line 22:
```suggestion
        title="The AYON Shotgrid user",
```

fabiaserra, line 22: How about
```suggestion
        title="AYON service user",
```
?
since it's not really SG specific?

Minkiu: Hmm, you could have multiple service users, and in here we need to pick the one that we use when setting up the services/ash through the `AYON_API_KEY` ; which after writing this maybe could lead to mismatch of things, say we use another service (unrelated to shotgrid) to create an entity cause they share service name, the transmitter would ignore it eve tho it shouldn't so, maybe I should enforce (or create) a user that matches the secret username for SG? Thoughts?

services/processor/processor/processor.py
Minkiu: I replaced it with another one, but yeah, will set it as debug

services/shotgrid_common/ayon_shotgrid_hub/update_from_ayon.py
```diff
@@ -252,6 +252,12 @@ def _create_sg_entity(
 
     logging.debug(f"Creating Shotgrid entity {sg_type} with data: {data}")
 
-    return sg_session.create(sg_type, data)
+    try:
```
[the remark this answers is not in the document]
Minkiu: If the entity already exists, it will raise, which will be handled by the `try...except` in line `68`, we would also exiting early if we just created a new one, since we are returning the newly created one.

server/settings/main.py
fabiaserra: yeah I'd say so, if it can happen, it will happen haha

services/shotgrid_common/ayon_shotgrid_hub/update_from_ayon.py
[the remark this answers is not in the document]
fabiaserra: got it

PR #67, opened 2024-01-22
Fix bug where custom project code field wasn't being used on sync from Ayon to SG

fabiaserra:
The custom project code field was missing on this call:
```
    # Append the project's direct children.
    for ay_project_child in entity_hub._entities_by_parent_id[entity_hub.project_name]:
        ay_entities_deck.append((
            get_sg_entity_as_ay_dict(sg_session, "Project", sg_project["id"], project_code_field),
            ay_project_child
        ))
```

Although perhaps better than querying the `sg_project` again the best would be to append the existing `sg_project` variable that already exists on the function and contains the project code field, something like:

```
    # Append the project's direct children.
    for ay_project_child in entity_hub._entities_by_parent_id[entity_hub.project_name]:
        ay_entities_deck.append((
           _sg_to_ay_dict(sg_project, project_code_field)
            ay_project_child
        ))
```

@Minkiu thoughts?

### DIFF
--- a/services/shotgrid_common/utils.py
+++ b/services/shotgrid_common/utils.py
@@ -415,7 +415,10 @@ def get_sg_entity_as_ay_dict(
     query_fields = list(SG_COMMON_ENTITY_FIELDS)
     if extra_fields and isinstance(extra_fields, list):
         query_fields.extend(extra_fields)
-
+    
+    if project_code_field not in query_fields:
+        query_fields.append(project_code_field)
+        
     sg_entity = sg_session.find_one(
         sg_type,
         filters=[["id", "is", sg_id]],
--- a/services/shotgrid_common/utils.py
+++ b/services/shotgrid_common/utils.py
@@ -43,8 +43,6 @@ def _sg_to_ay_dict(sg_entity: dict, project_code_field=None) -> dict:
 
         if not label and not task_type:
             raise ValueError(f"Unable to parse Task {sg_entity}")
-        else:
-            label = sg_entity["step"]["name"]
 
         name = slugify_string(label)
 
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,18 @@ Temporary Items
 # IDEA
 ######
 .idea/
+
+# Add symlinked files to ignore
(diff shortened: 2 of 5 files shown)


Minkiu:
> The custom project code field was missing on this call:
> 
> ```
>     # Append the project's direct children.
>     for ay_project_child in entity_hub._entities_by_parent_id[entity_hub.project_name]:
>         ay_entities_deck.append((
>             get_sg_entity_as_ay_dict(sg_session, "Project", sg_project["id"], project_code_field),
>             ay_project_child
>         ))
> ```
> 
> Although perhaps better than querying the `sg_project` again the best would be to append the existing `sg_project` variable that already exists on the function and contains the project code field, something like:
> 
> ```
>     # Append the project's direct children.
>     for ay_project_child in entity_hub._entities_by_parent_id[entity_hub.project_name]:
>         ay_entities_deck.append((
>            _sg_to_ay_dict(sg_project, project_code_field)
>             ay_project_child
>         ))
> ```
> 
> @Minkiu thoughts?

That would probably work, we'd need to change the function to remove the leading `_` but I get your point; although  `get_sg_entity_as_ay_dict` will add more fields to the query, while the of `sg_project` might not have everything required...

fabiaserra:
> That would probably work, we'd need to change the function to remove the leading `_` but I get your point; although `get_sg_entity_as_ay_dict` will add more fields to the query, while the of `sg_project` might not have everything required...

Yeah I had the same thought, let's leave it like this then

services/shotgrid_common/utils.py
```diff
@@ -43,8 +43,6 @@ def _sg_to_ay_dict(sg_entity: dict, project_code_field=None) -> dict:
 
         if not label and not task_type:
             raise ValueError(f"Unable to parse Task {sg_entity}")
-        else:
```
fabiaserra, line 46: @Minkiu this was overriding the label every time

Minkiu: Here I though I was playing it safe... :facepalm:  thanks!

Minkiu (approved):
Thanks for the contribution, merging!

PR #68, opened 2024-01-23
Expose whether to use SG local storage during publish

fabiaserra:
This exposes whether we want to try use the SG local storage setting during publish to split the relative path or not. This also exposes which of the entries of local storage to use as we actually had more than one defined and that's how I came across this error because it was trying to split the path with a root that didn't exist.

Screenshot 2024-01-23 at 4 51 38 PM

From my testing this doesn't seem necessary in Linux or our SG as by just passing the `local_path` to the published file it seems to automatically detect the local storage entry but this might come useful for others.


Minkiu (approved):
Looking good, thanks for the contribution, merging!

PR #73, opened 2024-02-13
Use 'AYONAddon' base class

iLLiCiTiT:
## Description
Use `AYONAddon` as base class for client addon.


Minkiu (approved):
Looks good, thanks!

PR #72, opened 2024-02-07
`settings` Add a map for entity's custom attributes

Minkiu:
## Description
Add two new settings to the AYON-Shotgrid integration,:
* "Shotgrid Enabled Entities": User can define which entities they have enabled in their instance. 
* "Folder Attributes Map": User can define which AYON folder attributes map to Shotgrid entity fields. i.e.: `clipIn` -> `sg_cut_in`.

image
(image originally posted by @fabiaserra at #79 )

Also removed some unnecessary logging, and modify the services to make use of the new settings, bug fixes and quality of life improvements.

## Credits
Big shoutout to @fabiaserra and @Minkiu for their awesome work on this PR! Couldn't have done it without them!

## Changelog
3b3b424 `services` Remove SG `"Version"` sync We do not want to synchronise, and this is delegated to the publishing step, to create it in both AYON and SG.
5a2a8a9 `server.settings` Add `ShotgridCompatibility` A new section in the Shotgrid settings, where two new settings have been  added:  * "Shotgrid Enabled Entities": User can define which entities they have     enabled in their instance.  * "Folder Attributes Map": User can define which AYON folder attributes     map to Shotgrid entity fields. i.e.: `clipIn` -> `sg_cut_in`.
a0a4b98 `shotgrid_common.utils` Follow EntityHub's struc
7afd7fa `services` Adapt to new EntityHub struct Modified all places where we were accessing information that now lives in either `attribs` or `data` in the resulting `dict` in the new `shotgrid_common.utils._sg_to_ay_dict` which follows the EntityHub structure more closely.
09e7883 `services` Honour "Folder Attributes Map" setting
9e4e73e `services` Honour "Shotgrid Enabled Entities" setting
4e36120 `ayon_shotgrid_hub` Fix when matching hierarchies
854510d `ayon_shotgrid_hub` Reduce checks before sync Remove a check for missing attributes when initializing the Hub, which was adding noise and potential confusion to the user.
8904556 `services.processor` Pass `ShotgridProcessor` instance to handlers We were passing a lot of arguments to the `processor/handlers` this changes simplifies things a bit by just sending the actual class, so we then access the necessary attributes when creating the `AyonShotgridHub`.

## Todo
- [ ] Documentation, this needs to include tutorial on how to configure the addon and anatomy preset > this PR
- [x] Establishing minimal defaults for Attribute mapping. Those has to be created on SG if they don't exists > this PR
- [x] Investigate update and sync statuses (entity hub doesn't seem to allow this yet?) > next PR

[...]

PR #75, opened 2024-03-28
Fixes for "`settings` Add a map for entity's custom attributes #72"

fabiaserra:
A few bug fixes on top of https://github.com/ynput/ayon-shotgrid/pull/72

* Duplicated tasks from SG fail to sync to Ayon and because of how the system is designed the whole thing blocks from syncing more things. This now avoids trying to create entities that have the same name under the same parent... although that's still not ideal as that will create an out of sync behavior with Shotgrid DB and won't allow us to update the task. Ideally Ayon wouldn't have the limitation of task names needing to be unique as I have said in the past also with products https://community.ynput.io/t/restriction-of-uniqueness-of-product-names/1232 ... but alternatively, we need to figure out a way to show the user that there's a duplicated task and they need to rename it in SG if they want it synced and not get weird behaviors...
* Add support to catch attrib_changed events from Ayon so they sync to SG

Other questions for @Ynput:

* Why are `status` and `tags` changes on the entities not firing any events?
* Custom attributes don't map to the same field on all entities in SG as some of those are built-in and some aren't and the non-built-in start with `sg_`  (i.e., `sg_start_date`)... whereas the built-in ones don't (i.e., `start_date`) so we would need to either check whether the optional `sg_` prefix exists or allow for setting different custom attributes per entity...

### DIFF
--- a/client/ayon_shotgrid/addon.py
+++ b/client/ayon_shotgrid/addon.py
@@ -48,7 +48,7 @@ class ShotgridAddon(OpenPypeModule, IPluginPaths):
     def create_shotgrid_session(self):
         from .lib import credentials
 
-        sg_username = os.getenv("AYON_SG_USERNAME")
+        sg_username = os.getenv("USER") or os.getenv("AYON_SG_USER")
         proxy = os.environ.get("HTTPS_PROXY", "").lstrip("https://")
 
         return credentials.create_sg_session(
--- a/client/ayon_shotgrid/plugins/publish/collect_shotgrid_session.py
+++ b/client/ayon_shotgrid/plugins/publish/collect_shotgrid_session.py
@@ -11,10 +11,10 @@ class CollectShotgridSession(pyblish.api.ContextPlugin):
     label = "Collecting Shotgrid session"
 
     def process(self, context):
-        user_login = os.getenv("AYON_SG_USERNAME")
+        user_login = os.getenv("USER") or os.getenv("AYON_SG_USER")
         if not user_login:
             raise KnownPublishError(
-                "Have you logged in into Ayon Tray > Shotgrid?"
+                "User not found in environment, make sure it's set."
             )
 
         shotgrid_module = context.data["openPypeModules"]["shotgrid"]
--- a/server/frontend/dist/index.html
+++ b/server/frontend/dist/index.html
@@ -31,7 +31,7 @@
                 <th>Code</th>
                 <th>In AYON?</th>
                 <th>In Shotgrid?</th>
-                <th>Syncronize</th>
+                <th>Synchronize</th>
             </tr>
         </thead>
         <tbody class="sg-addon-projects-table-body" id="sg-addon-projects-table-body">
--- a/server/frontend/dist/shotgrid-addon.js
+++ b/server/frontend/dist/shotgrid-addon.js
@@ -39,7 +39,7 @@ const init = () => {
 
 const populateTable = async () => {
   /* Get all the projects from AYON and Shotgrid, then populate the table with their info
-  and a button to Syncronize if they pass the requirements */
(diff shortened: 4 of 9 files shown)


BigRoy:
> * Why are `status` and `tags` changes on the entities not firing any events?

Would be for @martastain I think?

> Custom attributes don't map to the same field on all entities in SG as some of those are built-in and some aren't and the non-built-in start with sg_ (i.e., sg_start_date)... whereas the built-in ones don't (i.e., start_date) so we would need to either check whether the optional sg_ prefix exists or allow for setting different custom attributes per entity...

What's the question exactly? Which of the two to do? How to do it? Or?

fabiaserra:
> What's the question exactly? Which of the two to do? How to do it? Or?

Just opening the debate on how we would prefer to solve it, probably the easiest is to just check for attributes with both the `sg_` prefix and without... and documenting it to the user so they know they have to create all of those fields on the entities if they want to be able to sync Ayon <-> SG

fabiaserra:
> Current state is quite limited and this should be considered for future dev.

Honestly the status syncing from SG to Ayon was one of the reasons I was more excited to move to Ayon and I think it's crucial to sell the idea to the artists to use Ayon web server instead of the existing features in SG. Until that happens I don't see really any of my artists checking the Ayon portal over SG's.

> Yeah I am hitting the same problem. Ideal solution would be somehow populate all available SG attributes into settings enumerator (perhaps impossible at the moment @martastain, right?) so we avoid manual typing.

Yeah that's a good idea, it would make things easier and we could probably populate the attributes on the SG entities from Ayon directly instead of having to go through each entity and adding them manually in SG which is a big pain in the ass ATM.

> Also this conversion from ex. `Clip In` to `sg_clip_in` is very implicit and not easy to figure out. We should have the same manual input `Clip In` as it is in SG and convert it to `sg_clip_in` automatically.

Yeah 100%, the differences in attribute names makes this very confusing...

BigRoy:
> Current state is quite limited and this should be considered for future dev.

Can you elaborate _why_ that is? I would have definitely expected status changes on AYON to be something to be able to 'listen' to and react upon. Isn't that kind of the idea what the services were about. I'm pretty sure @martastain showed me an example to listen to status changes during the AYON summit hackathon.

So what in particular is _limited?_

client/ayon_shotgrid/addon.py
```diff
@@ -48,7 +48,7 @@ def get_local_storage_key(self):
     def create_shotgrid_session(self):
         from .lib import credentials
 
-        sg_username = os.getenv("AYON_SG_USERNAME")
+        sg_username = os.getenv("USER") or os.getenv("AYON_SG_USER")
```
jakubjezek001, line 51:
```suggestion
        sg_username = os.getenv("AYON_SG_USERNAME") or os.getenv("USER")
```

jakubjezek001: It seems that the order should be flipped, right? Also is there any particular reason for proposing the original variable name?

fabiaserra: either way, in our studio there's never a case where one would point to something else but yeah I can revert it and keep the original name of the env.

I think I renamed it just to follow closer to the original env used in OP `OPENPYPE_SG_USER`

fabiaserra: you already merged it so I guess you will do the changes?

jakubjezek001: yeah, sorry :D

fabiaserra: I think ideally we should be leveraging the existing login framework from Ayon... having multiple login dialogs is very cumbersome and there's really 0 need to have an authentication for SG as the whole plugin uses the API key to publish

PR #76, opened 2024-03-28
`server` Simple support to sync users from SG over to Ayon

fabiaserra:
Quick implementation to sync the user database from SG to Ayon. 

This also adds a change to the existing populate tables functionality to filter out all the projects that we wouldn't be able to sync (because they have whitespaces or not a code name) so the table isn't that long and it's easier to navigate through the projects... although ideally the whole server UI needs a revamp


jakubjezek001:
> This also adds a change to the existing populate tables functionality to filter out all the projects that we wouldn't be able to sync 

Before, we had the opportunity to visually identify the underlying issues within those specific projects and proactively address them. Now, without this visibility, valuable time may be spent attempting to pinpoint the root cause of any unexpected occurrences

server/frontend/dist/index.html
```diff
@@ -9,6 +9,7 @@
 
   <body>
     <h1>Shotgrid Sync</h1>
+    <button id="sync-users-button"  class="sync-users-button" onclick="syncUsers();">Sync Users</button>
```
jakubjezek001, line 12: - It would be beneficial to include details about the direction of sync.
- Additionally, what are the implications if a user has already been synced and subsequently leaves ShotGrid?
- Will they be removed from AY's database or simply archived?

server/frontend/dist/shotgrid-addon.js
```diff
@@ -160,12 +296,19 @@ const getShotgridProjects = async () => {
...
+        !project.attributes.name.includes(" ") &&
+        project.attributes[`${addonSettings.shotgrid_project_code_field}`]
+      ) {
+        sgProjectsConformed.push({
+          "name": project.attributes.name,
+          "code": project.attributes[`${addonSettings.shotgrid_project_code_field}`],
+          "shotgridId": project.id,
+          "ayonId": project.attributes.sg_ayon_id,
+        })
+      }
```
jakubjezek001, line 311: This should not be part of this PR please.

jakubjezek001 (changes requested):

fabiaserra:
> > This also adds a change to the existing populate tables functionality to filter out all the projects that we wouldn't be able to sync
> 
> Before, we had the opportunity to visually identify the underlying issues within those specific projects and proactively address them. Now, without this visibility, valuable time may be spent attempting to pinpoint the root cause of any unexpected occurrences

Fair, in my case I prefer it this way as otherwise there's lots of noise of template projects and other legacy things that I know for sure I won't ever sync and a table widget is already clunky enough for this workflow that the more filtering the better. Also the documentation that shows in the same page clearly states both requirements for the sync to work... 
image

I think I´d rather steal the UI from another of the addons that might have spend more time trying to make this friendlier. Can you show a screenshot of the addon-ftrack and addon-kitsu to compare how they handle it?

server/frontend/dist/shotgrid-addon.js
fabiaserra: sure

server/frontend/dist/index.html
fabiaserra: We don't delete users in SG, we simply mark them unactive, I could also add functionality to sync unactive users to keep both on sync... but we have so many unactive users I'd rather not do that. Ideally that would happen through the event system instead of a button so if an Ayon user gets marked unactive it gets propagated to SG and viceversa

Similarly as I mentioned in Discord yesterday, the next step is to add support for the users access/roles to grab that info from SG. Studios that have SG won't be doing the user access from Ayon, they will simply do it from SG so we should catch that event in the project entity and add the users list to the Ayon project

server/frontend/dist/shotgrid-addon.js
fabiaserra: done

fabiaserra:
> You you have a chance @fabiaserra please resolve conflicts and lets merge it ;)

Done, please respond to https://github.com/ynput/ayon-shotgrid/pull/86/files/25d1b7fdf244e229a21bb404ddbe48bea0192f13#r1565979186

BigRoy:
@jakubjezek001 the ball is in your court - can we merge this?

jakubjezek001:
Also @fabiaserra would you mind to merge latest develop into your branch? Thank you.

fabiaserra:
> Also @fabiaserra would you mind to merge latest develop into your branch? Thank you.

You did it no?

jakubjezek001:
> > Also @fabiaserra would you mind to merge latest develop into your branch? Thank you.
> 
> You did it no?

Yeah I have done it - didn't know I can do it myself ;)

jakubjezek001 (changes requested):

frontend/dist/shotgrid-addon.js
```diff
@@ -122,6 +122,142 @@ const populateTable = async () => {
...
+  })
+}
+
+
+const getShotgridUsers = async () => {
+  /* Query Shotgrid for all active users. */
+  const sgBaseUrl = `${addonSettings.shotgrid_server.replace(/\/+$/, '')}/api/v1`
+  sgAuthToken = await axios
+    .post(`${sgBaseUrl}/auth/access_token`, {
+        client_id: addonSettings.shotgrid_script_name,
```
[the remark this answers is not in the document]
fabiaserra: I actually have that update on my fork already from a bit ago but I forgot to update this branch as well sorry https://github.com/fabiaserra/ayon-shotgrid/blob/release/alkemyx/frontend/dist/shotgrid-addon.js#L150

frontend/dist/shotgrid-addon.js
```diff
@@ -122,6 +122,142 @@ const populateTable = async () => {
...
+      },
+      "password": login,
+    })
+    .then((result) => result)
+    .catch((error) => {
+      console.log("Unable to create user in AYON!")
+      console.log(error)
+      call_result_paragraph.innerHTML = `Unable to create user in AYON! ${error}`
+  });
+}
```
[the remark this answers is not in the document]
fabiaserra: hmm yeah I'm not sure, my syncs still work fine, I just synced a few more differences I had on our fork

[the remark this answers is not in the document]
jakubjezek001: So some changes needed to be done so we could merge this. Please review following PR on your fork @fabiaserra https://github.com/fabiaserra/ayon-shotgrid/pull/1. This should probably interfere with your user base since we are only normalizing users which are having email in login. Also I had removed the password and active flag since it is active by default. The original SG login was added into `user.data.sg_login`.

frontend/dist/shotgrid-addon.js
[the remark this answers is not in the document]
jakubjezek001: Thank you @fabiaserra

fabiaserra:
FYI I recently just finished adding accessGroups synchronization in our fork of `ayon-shotgrid`: https://github.com/fabiaserra/ayon-shotgrid/blob/release/alkemyx/frontend/dist/shotgrid-addon.js#L181 it has some assumptions that only would work in our studio at the moment such as the Flow groups being `admin`, `management`, `executive`, `lead` and `artist` but perhaps you can get inspiration on abstracting that away on the settings

kalisp:
I am wondering if we need this PR or we should just move to syncing users via regular event system. Having separate javascript logic just for that seems weird.

fabiaserra:
> I am wondering if we need this PR or we should just move to syncing users via regular event system. Having separate javascript logic just for that seems weird.

yeah as I said it still has a small use case for forcing the sync of users without the need for an event, in the case where users don't have a project assigned yet. It's very few cases where that could happen but it also helps for debugging purposes so it doesn't hurt to have it. On any case, I have both workflows implemented on my fork already

jakubjezek001 (approved):

PR #79, opened 2024-04-01
Custom attributes mapping with type and scope

fabiaserra:
## Description
This adds much better support to sync custom attributes by exposing types and scope in the settings so we can auto-populate the attributes on the entities for the user instead of having to add them manually:
image

It also abstracts away the "sg_" prefix on the attributes so it supports syncronization of both native and non-native SG fields as some entities have those fields as native and some others don't.

This also tackles some of the existing limitation of status/tags events not triggering, although the sync of statuses is currently only working from Ayon to SG because the Ayon API doesn't seem to update the status entity for some reason. I also fixed the status syncronization from SG to Ayon project so we aren't case sensitive.

@jakubjezek001 is there a plan to add support of defining statuses by entity instead of just a global list of entries?

## TODO:
* Make the type setting enumerator a single entry selection and not a list
* Add custom attributes during entities creation
* Fix status/tags synchronization

### DIFF
--- a/server/frontend/dist/shotgrid-addon.css
+++ b/server/frontend/dist/shotgrid-addon.css
@@ -23,52 +23,52 @@ select {
 }
 
 table {
-table-layout: fixed;
-width: 100%;
-border-collapse: collapse;
-border: 4px solid #2C313A;
-margin-top: 25px;
+  table-layout: fixed;
+  width: 100%;
+  border-collapse: collapse;
+  border: 4px solid #2C313A;
+  margin-top: 25px;
 }
 
 thead {
-text-align: left;
-background-color: #384956;
+  text-align: left;
+  background-color: #384956;
 }
 
 thead th:nth-child(1) {
-width: 25%;
+  width: 25%;
 }
 
 thead th:nth-child(2) {
-width: 25%;
+  width: 25%;
 }
 
 thead th:nth-child(3) {
-width: 10%;
+  width: 10%;
 }
 
 thead th:nth-child(4) {
-width: 10%;
+  width: 10%;
 }
 
 thead th:nth-child(5) {
-width: 30%;
+  width: 30%;
 }
 
 tbody td {
-text-align: left;
-border: 3px solid #2C313A;
+  text-align: left;
+  border: 3px solid #2C313A;
 }
 
 th,
 td {
-border: 3px solid #2C313A;
+  border: 3px solid #2C313A;
 }
 
 table tr:not(:first-child):nth-child(odd) {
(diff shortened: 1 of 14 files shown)


server/frontend/dist/shotgrid-addon.css
```diff
@@ -1,79 +1,74 @@
...
 
 th,
 td {
-  border: 3px solid #2C313A;
+border: 3px solid #2C313A;
 }
 
-tbody tr:nth-child(odd) {
-  background-color: var(--gray-50);
+table tr:not(:first-child):nth-child(odd) {
```
fabiaserra, line 68: this fixes the table style so each row background alternates color so it's much easier to read which project belongs to which row

jakubjezek001:
@fabiaserra man this is awesome, thanks a lot. I was about to do it and now I don't need to! :D I will go trough it later but for now , this is really solid way to do - talking about Type, Scope and ditching `sg_` prefix. I wanted to approach it my self the way that user could even copy paste the name (e.g. Cut In) from SG and paste it here in SG column. Since SG attributes are kind of implicit.

services/shotgrid_common/utils.py
```diff
@@ -184,9 +264,10 @@ def create_sg_entities_in_ay(
     project_entity.folder_types = new_folder_types
 
     # Add Shotgrid Statuses to Project Entity
-    for status in get_sg_statuses(sg_session):
-        status_short_name, status_name = status
-        project_entity.statuses.create(status_name, short_name=status_short_name)
+    ay_status_codes = [s.short_name.lower() for s in list(project_entity.statuses)]
+    for status_code, status_name in get_sg_statuses(sg_session).items():
+        if status_code.lower() not in ay_status_codes:
+            project_entity.statuses.create(status_name, short_name=status_code)
```
jakubjezek001, line 270: I like this a lot but somehow feel this should be done in separate PR. I believe that since we are dealing with attributes mapping, statuses and tags synchronization should be separated. I will do it during the merging process as well as it had been done before.

fabiaserra: not sure I follow, what do you mean it had been done before?

fabiaserra: I just pushed another big pass of the code to clean up a lot of the variables to be consistent across all the codebase so it's easier to follow

jakubjezek001:
> Syncing status and tags from AYON to Shotgrid is now working, it's only the other way around not working because of the AYON API not permitting it

Perhaps @iLLiCiTiT or @martastain can answer this question?

PR #80, opened 2024-04-05
Update to ayon_core

fabiaserra:
Update to ayon_core

More info: https://community.ynput.io/t/openpype-to-ayon-core-addon-conversion/1209/3


jakubjezek001:
This is already part of #83

PR #81, opened 2024-04-05
Remove use of SG secret to retrieve API key

fabiaserra:
Remove the use of the Ayon secret as non-admin users were getting an error trying to publish as reported here https://community.ynput.io/t/shotgrid-addon-secrets-permissions-and-proxy-problems/1343/7


jakubjezek001:
Thank you very much for this changes @fabiaserra. I will implement them within the scope of following PR https://github.com/ynput/ayon-shotgrid/pull/83

PR #86, opened 2024-04-10
Small bug fixes on latest develop

fabiaserra:
Some small fixes that I noticed on latest develop


services/shotgrid_common/utils.py
```diff
@@ -219,6 +219,7 @@ def create_ay_fields_in_sg_project(
             attribute containing the type of data and the scope of the
             attribute.
     """
+    # NOTE: this is not needed?
```
fabiaserra, line 222: @jakubjezek001 I noticed you added this, is this needed now? the `type` should always exist on this case since `SG_PROJECT_ATTRS` is our constant?

jakubjezek001: Yeah this was probably passed from some code migration. We definitely need it, you are right. Thanks

fabiaserra: we do or we do NOT?

jakubjezek001, line 222:
```suggestion
```

jakubjezek001 (approved):

services/shotgrid_common/utils.py
fabiaserra: wait, I added that `NOTE` myself to discuss whether the code following that was required. I believe that is not since `SG_PROJECT_ATTRS` includes ALL the types of the queries you are doing so you don't need to construct that `ayon_attribs_mapping`?

BigRoy: Pinging @jakubjezek001

PR #94, opened 2024-05-02
Service tools: Helper scripts to run services out of docker

iLLiCiTiT:
## Description
The tools are meant for easier development of services, especially when are required changes out of the service code itself (e.g. in dependencies).

### Additional information
New scripts do allow to run each service or all of them at once. To do so it is required to create env inside service tools directory.


jakubjezek001 (approved):
LGTM

PR #95, opened 2024-05-03
Update logging imports and error handling in various modules. Set up basic logging configuration in the main processor module. Remove redundant nxtools import statements.

jakubjezek001:


services/shotgrid_common/utils.py
```diff
@@ -14,11 +15,50 @@
...
+    ch.setFormatter(formatter)
+
+    # add ch to logger
+    logger.addHandler(ch)
+
+    return logger
+
+
+# create logger
+log = get_logger(__name__)
```
jakubjezek001, line 59: I am not so sure about correct place for this function and also if this is the best way to create the logger @iLLiCiTiT or @martastain you might have perhaps better idea?

services/shotgrid_common/utils.py
```diff
@@ -14,11 +15,50 @@
...
-from nxtools import logging
 import shotgun_api3
 
 
+def get_logger(name: str) -> logging.Logger:
+    """Return a logger instance with the given name."""
+    # get environment variable DEBUG level
+    log_level = os.environ.get("LOGLEVEL", "INFO").upper()
+
+    logger = logging.Logger(name)
```
iLLiCiTiT, line 28: Avoid reinitialization of the same logger...
```suggestion

_loggers = {}


def get_logger(name: str) -> logging.Logger:
    """Return a logger instance with the given name."""
    if name in _loggers:
        return _loggers[name]

    # get environment variable DEBUG level
    log_level = os.environ.get("LOGLEVEL", "INFO").upper()

    logger = logging.Logger(name)
    _loggers[name] = logger
```

services/shotgrid_common/utils.py
```diff
@@ -14,11 +15,50 @@
 from ayon_api.entity_hub import ProjectEntity
 from ayon_api.utils import slugify_string
 from ayon_api import get_attributes_for_type
+import logging
```
iLLiCiTiT, line 18: This import should be earlier...

iLLiCiTiT:
If it works then I'm fine with it...

PR #96, opened 2024-05-06
Fixing thumbnail uploading

jakubjezek001:
Adjusted thumbnail upload parameters to include version type and id for image products. Improved handling of published file types.

### DIFF
--- a/client/ayon_shotgrid/plugins/publish/integrate_shotgrid_publish.py
+++ b/client/ayon_shotgrid/plugins/publish/integrate_shotgrid_publish.py
@@ -181,9 +181,9 @@ class IntegrateShotgridPublish(pyblish.api.InstancePlugin):
 
             if instance.data["productType"] == "image":
                 sg_session.upload_thumbnail(
-                    sg_published_file["type"],
-                    sg_published_file["id"],
-                    local_path
+                    sg_published_file["version"]["type"],
+                    sg_published_file["version"]["id"],
+                    local_path,
                 )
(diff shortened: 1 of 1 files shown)


client/ayon_shotgrid/plugins/publish/integrate_shotgrid_publish.py
```diff
@@ -181,9 +181,9 @@ def process(self, instance):
 
             if instance.data["productType"] == "image":
```
fabiaserra, line 182: this feels honestly wrong that we are only doing this for `image` product types. We should be uploading thumbnail using the `thumbnail` representation on the instance? although thumbnails are now being automatically created for the SG Versions that get a video uploaded so maybe it wouldn't be needed for those. On any case, I'd expect that functionality to be on the `integrate_shotgrid_version.py` and not on this plugin

iLLiCiTiT: Agree, but that's an enhancement, whereas this should be bugfix.

fabiaserra: are you sure it's a bug fix? uploading the thumbnail to the published file directly instead of the version is not a bug, it was probably a choice of the developer. I think it's more of a bug now that the published files if there's an image it will replace the thumbnail of the version (which might already exist if there's an uploaded video)... why not just doing it right how you'd want it to behave?

PR #108, opened 2024-06-05
Update SG version entity data to include movie path even for thumbnails only 

jakubjezek001:
- Update data with movie and frames paths for ShotGrid integration.
- also fixing the issue with double version publishing with workfile product type  instances


client/ayon_shotgrid/plugins/publish/integrate_shotgrid_version.py
```diff
@@ -68,15 +68,21 @@ def process(self, instance):
                 # Replace the frame number with '%04d'
                 path_to_frame = re.sub(r"\.\d+\.", ".%04d.", local_path)
 
-                data_to_update["sg_path_to_frames"] = path_to_frame
+                data_to_update |= {
+                    "sg_path_to_movie": path_to_frame,
```
fabiaserra, line 72: why? this could be replacing the `sg_path_to_movie` from the video representation if the iteration happens after?

PR #109, opened 2024-06-06
Remove project code field being disabled on settings

fabiaserra:
### Changes

Makes the project code an editable field again in studio settings: `shotgrid_project_code_field`.
The default value is `code`

This setting then defines which field in the connected Shotgrid instance defines the project code to be synced.

### Additional info

This was reverted/removed in a recent PR https://github.com/ynput/ayon-shotgrid/pull/72 but being able to change the field that represents the project code on the Shotgrid instance is important because not all studios ShotGrid have `code` as the field used for setting the `id` as shown in this image of our Shotgrid:
image

If we aren't able to edit the project code field to `sg_code`, our syncs can't work.

### Testing notes

**Caveats**

Someone editing the setting to something else can break their syncs as the projects won't be found. However, similar to other settings in any other addon, a change implicates something.

### DIFF
--- a/server/settings/main.py
+++ b/server/settings/main.py
@@ -218,13 +218,13 @@ class ShotgridSettings(BaseSettingsModel):
     shotgrid_project_code_field: str = SettingsField(
         default="code",
         title="ShotGrid Project Code field name",
-        disabled=True,
         description=(
             "In order to create AYON projects, we need a Project Code, you "
             "can specify here which field in the ShotGrid Project "
             "entity represents it."
         ),
-        example="sg_code"
+        example="sg_code",
+        scope=["studio"],
     )
     enable_shotgrid_local_storage: bool = SettingsField(
         default=True,


m-u-r-p-h-y (commented):
This PR description is missing a lot of information.

Instead of exclamation mark it should clearly explain the reason why this field should be editable, as it is potentially breaking all running projects.

Some testing steps

jakubjezek001:
Consider eliminating the field entirely from the settings. I propose setting the default value to code, with an environment variable capable of overriding it, such as sg_code. This approach ensures flexibility while maintaining a default that prevents accidental alterations.

fabiaserra:
The reason I created this PR like this is because I thought it was obvious it had to be and it looked like a regression mistake, my bad assuming that was the case, I didn't know until your comments. As I said on the Discord channel: https://discord.com/channels/517362899170230292/563751989075378201/1254725003581194292 this was originally exposed on the settings because it was needed to be something studios can edit what field is used for project code fields. In our SG, it's always been `sg_code` for some reason (we don't even have `code` in our SG project entity) so it felt obvious at the time when Oscar was writing this addon to expose that setting and pass it along all the functions that used it to be able to use one or another
image

> Consider eliminating the field entirely from the settings. I propose setting the default value to code, with an environment variable capable of overriding it, such as sg_code. This approach ensures flexibility while maintaining a default that prevents accidental alterations.

What's your worry about making it editable? Any change on the settings can cause accidental problems really. What you are suggesting means removing all the changes that already exist in the codebase that use `project_code_field` (https://github.com/search?q=repo%3Aynput%2Fayon-shotgrid+project_code_field&type=code) and refactoring it so it's only an environment variable. That's a lot of work compared to just making this field editable. Has it caused any issues at any of your clients? What about just making the docstring more eye catching and warning the users that this is only for advanced cases?

BigRoy:
> Consider eliminating the field entirely from the settings. I propose setting the default value to code, with an environment variable capable of overriding it, such as sg_code. This approach ensures flexibility while maintaining a default that prevents accidental alterations.

I have a feeling this may not suffice once we get into the realm of looking into supporting multiple Shotgrid instances (different ones per project) to connect to from a single AYON server. @jakubjezek001 would that be problematic? Also, what's problematic about having it adjustable in AYON settings?

---

@fabiaserra could you take the time to just write up in the PR description what the change does in practice and _why_ you feel it's needed. I've taken the liberty to adjust your PR description in preparation - but it needs some of your knowledge to complete.

fabiaserra:
> @fabiaserra could you take the time to just write up in the PR description what the change does in practice and _why_ you feel it's needed. I've taken the liberty to adjust your PR description in preparation - but it needs some of your knowledge to complete.

I just updated it, let me know if it's still unclear

mkolar:
I generally agree that due to SG flexibility, this should be probably supported.

mkolar (approved):

PR #110, opened 2024-06-06
Fix proxy in session

fabiaserra:
Pass the right kwarg to shotgun session for proxy and use `re.sub` instead of string replacement as string replacement replaces any of the existing chars.


jakubjezek001 (approved):
Looking good, thank you ❤️

PR #83, opened 2024-04-06
Reimplementation of Tray user logging 

jakubjezek001:
## Changelog Description
We are reintroducing the functionality for users to access the ShotGrid addon via the tray widget. Previously, we had transitioned to an environment variable approach; however, following careful consideration of user input, we have opted to introduce two additional login methods. Users now have the option to log in using their username and password, mirroring the original process. Additionally, a novel method has been implemented where users can log in solely with their username, subsequently connecting to the ShotGrid Script API through studio settings. This approach involves a dedicated script with tailored permissions, catering specifically to fundamental user publishing requirements.

## Additional info
This PR is also solving several bugs caused by transition from OpenPype to ayon-core. Versoins are integrated again.

## Testing notes
Yeah this is long one! There are three ways to test this PR and we need to test all of them.

### Preparation for testing
All I will be describing in Dev mode ON
1. Since server settings were upgraded you need to update addon at your `ayon-docker/addons` folder. In repository root create and copy addon package to your local `ayon-docker` (this is still old structure so you should not clone this repo to addons folder directly). Or do it with the ZIP file uploading if you want to waist your time 😄 , or docker with server is not running from local machine. 
```
python .\create_package.py --skip-zip --output [PATH TO YOUR REPO]\ayon-docker\addon
```
2. Restart your server and then configure your bundle to lead to your client folder. 
image
3. go to addon settings `ayon+settings://shotgrid/shotgrid_server` and add your server. (you can also set env `HTTPS_PROXY`)
4. in **core** addon go to `ayon+settings://core/publish/ExtractReview/profiles/0/outputs`, find h264 preset and add **shotgridreview** to tags - this will upload publishing version to your ShotGrid server.
5. Then follow other steps which are related to particular user login methods

### Configuring Local storage
You will need to have configured your ShotGrid server > Site settings > File Management before the publishing.
Since default settings are enabled and set to `primary` you need to make sure this is also correctly configured on your server.
Read more about it [here ](https://help.autodesk.com/view/SGSUB/ENU/?guid=SG_Administrator_ar_data_management_ar_linking_local_files_html)

### Publishing test steps
1. Make sure your Local storage settings are in sync with ShotGrid File Management presets. Look above for more info. **Configuring Local storage**.
2. Then open any host or Traypublisher and publish some sequence of video with **review** activated
3. All plugins should work well and version should be published to SG as expected

- [x] default Env method
[...]

PR #112, opened 2024-06-06
Bring back changes so 'admin' roles can access all projects

fabiaserra:
I don't know why this was removed at some point, probably a bad merge conflict

### DIFF
--- a/client/ayon_shotgrid/plugins/publish/validate_shotgrid_user.py
+++ b/client/ayon_shotgrid/plugins/publish/validate_shotgrid_user.py
@@ -25,18 +25,38 @@ class ValidateShotgridUser(pyblish.api.ContextPlugin):
 
         sg_user = sg_session.find_one(
             "HumanUser",
-            [
-                ["login", "is", user_login],
-                ["projects", "name_contains", project_name]
-            ],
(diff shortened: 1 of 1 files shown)


fabiaserra:
> Include the link to the previously missing code. Additionally, the brief description lacks utility for the community. A more impactful approach would involve explaining the rationale or purpose behind the enhancements made to the codebase. Detailing new features and providing testing notes will benefit other contributors who may be curious about certain coding decisions. The current description falls short of being informative.

It was reverted by you here: https://github.com/ynput/ayon-shotgrid/pull/83/files#diff-3499dbb85188a04b0547eeb83635a87600df9c1a611ac2032ce6c571499fe9a7 and I don't see any explanation as to why that was done so I assume it was just a bad merge conflict because the original code had the feature that was originally already approved (https://github.com/ynput/ayon-shotgrid/pull/58) which is to allow `admin` users to write to any project

The changes are simple, it's just bringing back that functionality so `admin` users can access all projects.

# Testing

1. Using an SG user that has `admin` role try to publish to an SG project that the user is not a participant

jakubjezek001 (approved):
Great! thank you

PR #114, opened 2024-06-06
Ignore events by default unless it's a supported event by non-api user

fabiaserra:
There was a bug where SG events weren't being ignored by default, the default should be to ignore, otherwise user API events were causing a recursion of updates (and failing to do so)

### DIFF
--- a/services/leecher/leecher/listener.py
+++ b/services/leecher/leecher/listener.py
@@ -236,7 +236,7 @@ class ShotgridListener:
                     if not event:
                         continue
 
-                    ignore_event = False
+                    ignore_event = True
                     last_event_id = event["id"]
 
                     if (


jakubjezek001:
Thank you for highlighting the issue. It appears that the API events were encountering errors during user identification.

Could you confirm please that this was returning false even the event was emited from API?
https://github.com/fabiaserra/ayon-shotgrid/blob/c891fbbc9cc45046ae1fea3503c6b7498a04f7fd/services/leecher/leecher/listener.py#L271

fabiaserra:
> It appears that the API events were encountering errors during user identification.

Not sure I understand what you mean. 

> Could you confirm please that this was returning false even the event was emited from API? https://github.com/fabiaserra/ayon-shotgrid/blob/c891fbbc9cc45046ae1fea3503c6b7498a04f7fd/services/leecher/leecher/listener.py#L271

Yeah it returned `False` but if you look at the code, anything that's not catching the first conditional or the type is one of the `supported_event_types` it would not be ignored when it should be?
```python
                    if (
                        event["event_type"].endswith("_Change")
                        and event["attribute_name"].replace("sg_", "")
                        not in self.custom_sg_attribs
                    ):
                        # events related to custom attributes changes
                        # check if event was caused by api user
                        ignore_event = self._is_api_user_event(event)
                    elif event["event_type"] in supported_event_types:
                        # events related to changes in entities we track
                        # check if event was caused by api user
                        ignore_event = self._is_api_user_event(event)
```

kalisp (approved):
Seems fine.

That logic is due for refactoring, but that could be done in another PR.

PR #115, opened 2024-07-04
Add shotgrid user to version

antirotor:
Version was missing shotgrid user attached to it. This is finding logged user id and appending it to the version data.


jakubjezek001 (approved):
LGTM

PR #117, opened 2024-07-19
Refactor code to include new event hash generation function

jakubjezek001:
- Created new function `get_event_hash` in utils.py to generate SHA-256 hash from event topic and ID
- Added import for `get_event_hash` in listener.py
- Updated usage of `get_event_hash` in ShotgridListener class methods
- dispathing leecher events with summary with sg_event_id instead of adding it into ayon event hash
- Removing redundant source event update in processor
- Improving proc event description to better reflect on source SG event ID
- Leecher events are displaying SG payload in the message body

### DIFF
--- a/services/leecher/leecher/listener.py
+++ b/services/leecher/leecher/listener.py
@@ -6,6 +6,7 @@ Shotgrid and converts them to Ayon events, and can be configured from the Ayon
 Addon settings page.
 """
 import sys
+import json
 import time
 import signal
 import socket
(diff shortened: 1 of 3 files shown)


services/leecher/leecher/listener.py
```diff
@@ -183,7 +189,14 @@ def _find_last_event_id(self):
...
         for node in data["events"]["edges"]:
-            return node["node"]["hash"]
+            summary = node["node"]["summary"]
+            summary_data = json.loads(summary)
+            # TODO: remove hash in future since it is only used
+            #       as backward compatibility
+            if summary_data.get("sg_event_id"):
+                return node["node"]["summary"]["sg_event_id"]
+            else:
+                return int(node["node"]["hash"])
```
iLLiCiTiT, line 199:
```suggestion
                return summary_data["sg_event_id"]
            return int(node["node"]["hash"])
```

iLLiCiTiT (approved):
Changes make sense.

PR #118, opened 2024-07-25
Get user name based on login type

### LINKED ISSUE #119: Shotgrid session collection doesn't get user correctly
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

Issue reported in https://discord.com/channels/517362899170230292/563751989075378201/1266079526283116655

When publishing to Flow/SG, I noticed that it publishes as "ayon_user_publishing 1.0" and not the user, and it seems to only be publishing a still frame not the video file the farm generated.

### Expected Behavior:

shot grid should get the correct user name

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. change login type
2. publish a review from any host

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

### Relevant log output:

_No response_

### Additional context:

_No response_

MustafaJafar:
## Changelog Description
Issue reported on discord where collect shotgrid session doesn't respect the login type.
fixes #119 

## Testing notes:
1. change login type
2. publish a review from any host using different login types
    - [ ] env local
    - [ ]  env farm
    - [ ] tray_pass local
    - [ ] tray_pass farm
    - [ ]  tray api key local
    - [ ]  tray api key farm

### DIFF
--- a/client/ayon_shotgrid/addon.py
+++ b/client/ayon_shotgrid/addon.py
@@ -84,12 +84,7 @@ class ShotgridAddon(AYONAddon, ITrayAddon, IPluginPaths):
         from .lib import credentials
 
         if self._client_login_type == "env":
-            user_login = (
-                os.getenv("AYON_SG_USERNAME")
-                # TODO: Remove USER env variable in future once ayon-core deadline
-                # passing of AYON_SG_USERNAME is solved
-                or os.getenv("USER")
-            )
+            user_login = os.getenv("AYON_SG_USERNAME")
             sg_password = None
 
         else:
--- a/client/ayon_shotgrid/addon.py
+++ b/client/ayon_shotgrid/addon.py
@@ -80,6 +80,23 @@ class ShotgridAddon(AYONAddon, ITrayAddon, IPluginPaths):
     def get_local_storage_key(self):
         return self._local_storage_key or None
 
+    def get_credentials(self):
+        from .lib import credentials
+
+        if self._client_login_type == "env":
+            user_login = (
+                os.getenv("AYON_SG_USERNAME")
+                # TODO: Remove USER env variable in future once ayon-core deadline
+                # passing of AYON_SG_USERNAME is solved
+                or os.getenv("USER")
+            )
+            sg_password = None
+
+        else:
+            user_login, sg_password = credentials.get_local_login()
+        
+        return user_login, sg_password 
+
     def create_shotgrid_session(self):
         from .lib import credentials
         kwargs = {
@@ -89,36 +106,24 @@ class ShotgridAddon(AYONAddon, ITrayAddon, IPluginPaths):
         proxy = re.sub(r"https?://", "", os.environ.get("HTTPS_PROXY", ""))
         if proxy:
             kwargs["proxy"] = proxy
+        
+        sg_username, sg_password = self.get_credentials()
 
-        if self._client_login_type == "env":
-            sg_username = (
-                os.getenv("AYON_SG_USERNAME")
-                # TODO: Remove USER env variable in future once ayon-core deadline
-                # passing of AYON_SG_USERNAME is solved
-                or os.getenv("USER")
-            )
+        if self._client_login_type in {"env", "tray_api_key"}:
+            
             kwargs.update({
                 "username": sg_username,
                 "api_key": self._shotgrid_api_key,
                 "script_name": self._shotgrid_script_name,
             })
         elif self._client_login_type == "tray_pass":
-            sg_username, sg_password = credentials.get_local_login()
-
+           
             if not sg_username or not sg_password:
                 return None
 
             kwargs.update({
                 "username": sg_username,
                 "password": sg_password
-            })
-
-        elif self._client_login_type == "tray_api_key":
-            sg_username, _ = credentials.get_local_login()
-            kwargs.update({
-                "username": sg_username,
-                "api_key": self._shotgrid_api_key,
-                "script_name": self._shotgrid_script_name,
             })
 
         return credentials.create_sg_session(**kwargs)
--- a/client/ayon_shotgrid/addon.py
+++ b/client/ayon_shotgrid/addon.py
@@ -83,7 +83,7 @@ class ShotgridAddon(AYONAddon, ITrayAddon, IPluginPaths):
     def get_credentials(self):
         from .lib import credentials
 
-        if self._client_login_type == "env":
+        if self._client_login_type == "env" or os.getenv("HEADLESS_PUBLISH"):
             user_login = os.getenv("AYON_SG_USERNAME")
             sg_password = None
 
--- a/client/ayon_shotgrid/plugins/publish/collect_shotgrid_session.py
+++ b/client/ayon_shotgrid/plugins/publish/collect_shotgrid_session.py
@@ -13,10 +13,7 @@ class CollectShotgridSession(pyblish.api.ContextPlugin):
     client_login_type = "env"
 
     def process(self, context):
-        addons_manager = context.data["ayonAddonsManager"]
-        shotgrid_addon = addons_manager["shotgrid"]
-
-        user_login, _ = shotgrid_addon.get_credentials()
+        user_login = os.environ.get("AYON_SG_USERNAME")
 
         self.log.info(f"User login: {user_login}")
         if not user_login:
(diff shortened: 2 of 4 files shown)


fabiaserra (commented):
looks good!

client/ayon_shotgrid/addon.py
```diff
@@ -80,6 +80,23 @@ def is_local_storage_enabled(self):
...
 
+    def get_credentials(self):
+        from .lib import credentials
+
+        if self._client_login_type == "env":
+            user_login = (
+                os.getenv("AYON_SG_USERNAME")
+                # TODO: Remove USER env variable in future once ayon-core deadline
+                # passing of AYON_SG_USERNAME is solved
+                or os.getenv("USER")
```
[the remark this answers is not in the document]
BigRoy: How do we confirm whether that works? Then we can add it to the testing notes?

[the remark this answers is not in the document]
fabiaserra: it would be a matter of testing the three different types of login and confirm they work for both locally and the farm publishes

client/ayon_shotgrid/addon.py
```diff
@@ -80,6 +80,23 @@ def is_local_storage_enabled(self):
...
+            )
```
BigRoy, line 92: According to @fabiaserra [here](https://github.com/ynput/ayon-shotgrid/pull/118/files#r1692014726) we can remove the fallback.
```suggestion
            user_login = os.getenv("AYON_SG_USERNAME")
```

We'd have to need to update the TESTING NOTES to include:

1. Test with all different types of login options - they must work locally and on farm publishes.
    - [ ] env local
    - [ ] env farm
    - [ ] tray api key local
    - [ ] tray api key farm

(not sure if those are the actual login options, so consider it just an example of what to add)

MustafaJafar: @jakubjezek001  could confirm if the mentioned issue in the todo was solved ?

jakubjezek001: Yeah it is perhaps possible, since we are already passing `AYON_SG_USERNAME` in deadline.

fabiaserra: is that how the tray login is passing the username as well? I wonder why Steve on Discord mentioned his publish didn't work as the `AYON_SG_USERNAME` should have been set in the Deadline task and also collected?

client/ayon_shotgrid/addon.py
```diff
@@ -113,14 +121,6 @@ def create_shotgrid_session(self):
...
             })
 
-        elif self._client_login_type == "tray_api_key":
-            sg_username, _ = credentials.get_local_login()
-            kwargs.update({
-                "username": sg_username,
-                "api_key": self._shotgrid_api_key,
-                "script_name": self._shotgrid_script_name,
-            })
-
```
jakubjezek001, line 123: is there any reason why this had been removed?

This is now resulting in following error

```
  File "C:\Users\jakub\AppData\Local\Ynput\AYON\dependency_packages\ayon_2407261707_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "C:\CODE\__PYPE\ayon-shotgrid\client\ayon_shotgrid\plugins\publish\collect_shotgrid_session.py", line 23, in process
    raise KnownPublishError(
ayon_core.pipeline.publish.publish_plugins.KnownPublishError: User not found in environment, make sure it's set.
```

client/ayon_shotgrid/addon.py
```diff
@@ -80,6 +80,18 @@ def is_local_storage_enabled(self):
     def get_local_storage_key(self):
         return self._local_storage_key or None
 
+    def get_credentials(self):
+        from .lib import credentials
+
+        if self._client_login_type == "env":
+            user_login = os.getenv("AYON_SG_USERNAME")
```
MustafaJafar, line 87:
```suggestion 
            user_login = os.getenv("AYON_SG_USERNAME") or os.getenv("USER")
```

client/ayon_shotgrid/addon.py
MustafaJafar: Sorry, I missed this comment. 

> is there any reason why this had been removed?

It's not removed, it was moved. 
The `get_credentials` should return the username and password based on the login type. 
and the `create_shotgrid_session` should update the `kwargs` variable properly based on the login type. 

I wonder if this issue is related to the fallback that was removed. should we put it back 
https://github.com/ynput/ayon-shotgrid/pull/118/files#r1724742824 ?

client/ayon_shotgrid/addon.py
MustafaJafar: ehm.. like this ? 

```suggestion 
            user_login = os.getenv("AYON_SG_USERNAME") or os.getenv("AYON_USERNAME")
```

fabiaserra: no, I'm saying that we should check which env var is the one is being set across the different addons for logging in. I believe it might be we are just setting `AYON_USERNAME` now instead of `AYON_SG_USERNAME`. It would make more sense it's `AYON_USERNAME`

MustafaJafar: @iLLiCiTiT could you tell me which env var is recommended to get the current user name ?

fabiaserra: @MustafaJafar it should be pretty easy to check, just check the env vars that are present in your environment from any DCC and also check the env vars that are passed to Deadline when you submit a job

MustafaJafar:
```python
from ayon_core.lib import get_ayon_username
```

```suggestion 
            user_login = get_ayon_username()
```

jakubjezek001 (approved):
LGTM

client/ayon_shotgrid/addon.py
fabiaserra: you haven't applied this change yet have you?

MustafaJafar: As far as I know, this change will get the ayon user name which is not the same as sg name (login mail).
Also, we gave this PR a test run and faced an issue https://github.com/ynput/ayon-shotgrid/pull/118#issuecomment-2382987181

fabiaserra: oh, for us the sg name is the same as the ayon user name, we don't use the mail as `login` in SG, that's why there's a separate `email` field:
Screenshot 2024-10-02 at 9 28 12 AM

MustafaJafar: I think my ayon user name is not the same as SG name. I don't have them synced and also because I'm using an AYON instance for dev, I mostly use the default `Admin` username 
image

fabiaserra: yeah it doesn't look like, but you could change that `login` field on SG to match if you wanted

fabiaserra: I feel like for people who's SG set up doesn't have that on sync it doesn't make sense to use this login type so it should be documented the different types of logins and when to use one or another based on your set up

client/ayon_shotgrid/addon.py
```diff
@@ -80,6 +80,18 @@ def is_local_storage_enabled(self):
     def get_local_storage_key(self):
         return self._local_storage_key or None
 
+    def get_credentials(self):
+        from .lib import credentials
+
+        if self._client_login_type == "env":
```
kalisp, line 86: I think we are missing here check, if we are on farm. If we are, we should depend on `AYON_SG_USERNAME` env var as that is filled by Shotgrid AND is passed via `submit_publish_job`.

We need just double check in local publishing that `AYON_SG_USERNAME` contains correct value.

We might use check against `HEADLESS_PUBLISH` env var to differentiate, that we are on farm right now.

kalisp: I added check against `HEADLESS_PUBLISH`, tested, it seems to be working.

kalisp (approved):
Tested, it works.

It depends on `AYON_SG_USERNAME` being pushed to DL, so it cannot be removed during refactoring!

client/ayon_shotgrid/addon.py
MustafaJafar: marking this as resolved as mentioned in https://github.com/ynput/ayon-shotgrid/pull/118#pullrequestreview-2415867311

kalisp:
Couldnt reproduce, this solution wouldnt be working in Nuke as there is an issue with `pywintypes`.
Experimented with local and farm publishes, from AE and Nuke.
chrome_OwRnPpsTwk

There might be slight inconsistency if AYON user is not matching `AYON_SG_USERNAME` as first one is used in burnins, second one as user in SG. But that is totally unrelated to this issue

Could be reopened if still encountered.

PR #121, opened 2024-08-08
Enhancement: integration supports project level tag synchronization

kalisp:
## Changelog Description
Previously synchronization process failed on project tags. Now it check tags on Shotgrid/Flow project and adds them to Project Anatomy.

## Testing notes:

1. add some tags to SG project (edit value in Tags column, might need to be added to actually show the column)
image
image
2. create new AYON project with same name as SG, follow https://ayon.ynput.io/docs/addon_flow_admin#anatomy-preset-configuration
3. project could be synced manually in Shotgrid tab
image


MilaKudr (approved):
Screenshot 2024-08-12 162633
Screenshot 2024-08-12 162652

PR #122, opened 2024-08-13
Shotgrid: reviving archived assets is broken

kalisp:
## Changelog Description
This PR supports recreation of children Tasks in AYON if deleted Asset is revived in Shotgrid.

## Additional info
I don't like this solution too much, open for comments.

Main issue is that revival of assets on SG sends these events (in order) - `attribute_change` for CHILD task, `attribute_change` for Asset, `revival_entity`.
If we are processing first `attribute_change`, it wants to update non existing Task under non existing Asset.
Right now this throws new `MissingParentError` exception, that could be handled specifically. 

It fails processing event, it creates new source event on `shotgrid.event` (there is no way how to create new event on processing topic, imho), which gets put on the end of the queue. 
Then asset gets recreated, then this new event gets reprocessed (only once), this time hopefully it finds parent asset in AYON, so it finishes successfully (or fails forever).

## Testing notes:

1. create asset and task(s) in AYON
2. let them be synchronized to SG (easiest to run in this repo `service_tools/manage.ps` services`)
3. delete asset in AYON Editor
4. let it delete asset in SG
5. go to trash (top right your profile, Trash at the bottom, Asset)
image
6. check if asset and tasks under it got recreated in AYON

### DIFF
--- a/services/leecher/leecher/listener.py
+++ b/services/leecher/leecher/listener.py
@@ -6,6 +6,7 @@ Shotgrid and converts them to Ayon events, and can be configured from the Ayon
 Addon settings page.
 """
 import sys
+import json
 import time
 import signal
 import socket
@@ -13,7 +14,10 @@ import traceback
 from typing import Any
 from pprint import pformat
 
-from utils import get_logger
+from utils import (
+    get_logger,
+    get_event_hash,
+)
 
 from constants import (
     SG_EVENT_TYPES,
@@ -23,11 +27,13 @@ from constants import (
 import ayon_api
 import shotgun_api3
 
+# TODO: remove hash in future since it is only used as backward compatibility
 LAST_EVENT_QUERY = """query LastShotgridEvent($eventTopic: String!) {
   events(last: 1, topics: [$eventTopic]) {
     edges {
       node {
         hash
+        summary
       }
     }
   }
@@ -183,7 +189,16 @@ class ShotgridListener:
             return None
         data = response.data["data"]
         for node in data["events"]["edges"]:
-            return node["node"]["hash"]
+            summary = node["node"]["summary"]
+            summary_data = json.loads(summary)
+            # TODO: remove hash in future since it is only used
+            #       as backward compatibility
+            if summary_data.get("sg_event_id"):
+                return summary_data["sg_event_id"]
+
+            # return it old way
+            # TODO: remove in future
+            return int(node["node"]["hash"])
         return None
 
     def _get_last_event_processed(self, sg_filters):
@@ -205,7 +220,7 @@ class ShotgridListener:
             )
             last_event_id = last_event["id"]
 
-        return int(last_event_id)
+        return last_event_id
 
     def start_listening(self):
         """Main loop querying the Shotgrid database for new events
@@ -333,11 +348,16 @@ class ShotgridListener:
         Args:
             payload (dict): The Event data.
         """
-        description = f"Leeched {payload['event_type']}"
+        payload_id = payload["id"]
+        payload_type = payload["event_type"]
+        description = f"Leeched '{payload_type}' event  with ID '{payload_id}'"
         user_name = payload.get("user", {}).get("name", "Undefined")
 
         if user_name:
-            description = f"Leeched {payload['event_type']} by {user_name}"
+            description = (
+                f"Leeched '{payload_type}' event  with ID '{payload_id}' "
+                f"by '{user_name}'"
+            )
 
         # fix non serializable datetime
         payload["created_at"] = payload["created_at"].isoformat()
@@ -351,16 +371,20 @@ class ShotgridListener:
             project_id = payload.get("project", {}).get("id", "Undefined")
 
         sg_project = sg_projects_by_id[project_id]
+        new_event_hash = get_event_hash("shotgrid.event", payload["id"])
 
         ayon_api.dispatch_event(
             "shotgrid.event",
             sender=socket.gethostname(),
-            event_hash=payload["id"],
+            event_hash=new_event_hash,
             project_name=project_name,
             username=user_name,
             description=description,
-            summary=None,
+            summary={
+                "sg_event_id": payload_id,
+            },
             payload={
+                "message": json.dumps(payload, indent=2),
                 "action": "shotgrid-event",
                 "user_name": user_name,
                 "project_name": project_name,
--- a/services/leecher/leecher/listener.py
+++ b/services/leecher/leecher/listener.py
@@ -29,7 +29,7 @@ import shotgun_api3
 
 # TODO: remove hash in future since it is only used as backward compatibility
 LAST_EVENT_QUERY = """query LastShotgridEvent($eventTopic: String!) {
-  events(last: 20, topics: [$eventTopic]) {  
+  events(last: 20, topics: [$eventTopic]) {
     edges {
       node {
         hash
--- a/services/leecher/leecher/listener.py
+++ b/services/leecher/leecher/listener.py
@@ -400,7 +400,7 @@ class ShotgridListener:
             },
         )
 
-        self.log.info("Dispatched AYON event with payload:", payload)
+        self.log.debug(f"Dispatched AYON event with payload:{payload}")
 
 
 def service_main():
--- a/services/leecher/leecher/listener.py
+++ b/services/leecher/leecher/listener.py
@@ -1,8 +1,8 @@
 """
-A Shotgrid Events listener leecher for Ayon.
+A Shotgrid Events listener leecher for AYON.
 
 This service will continually run and query the EventLogEntry table from
-Shotgrid and converts them to Ayon events, and can be configured from the Ayon
+Shotgrid and converts them to AYON events, and can be configured from the AYON
 Addon settings page.
 """
 import sys
@@ -45,7 +45,7 @@ class ShotgridListener:
     log = get_logger(__file__)
 
     def __init__(self):
-        """Ensure both Ayon and Shotgrid connections are available.
+        """Ensure both AYON and Shotgrid connections are available.
 
         Set up common needed attributes and handle shotgrid connection
         closure via signal handlers.
@@ -232,7 +232,7 @@ class ShotgridListener:
         """Main loop querying the Shotgrid database for new events
 
         Since Shotgrid does not have an event hub per se, we need to query
-        the "EventLogEntry table and send these as Ayon events for processing.
+        the "EventLogEntry table and send these as AYON events for processing.
 
         We try to continue from the last Event processed by the leecher, if
         none is found we start at the moment in time.
@@ -293,7 +293,7 @@ class ShotgridListener:
                     if not event:
                         continue
 
-                    ignore_event = False
+                    ignore_event = True
                     last_event_id = event["id"]
 
                     if (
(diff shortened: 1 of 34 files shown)


MilaKudr (approved):
This PR works for me.

services/leecher/leecher/listener.py
```diff
@@ -28,12 +28,16 @@
 import shotgun_api3
 
 # TODO: remove hash in future since it is only used as backward compatibility
+# filters on only real events, starting with `Leeched`, not on `Recreated`
+# which have lower SG ID as they were tried, failed and should be retried
+# TODO filtering by 'leeched' or 'Leeched" doesnt work, by 'eeched' yes, wtf
 LAST_EVENT_QUERY = """query LastShotgridEvent($eventTopic: String!) {
-  events(last: 20, topics: [$eventTopic]) {
+  events(last: 3, topics: [$eventTopic], filter: "eeched") {
```
iLLiCiTiT, line 35:
```suggestion
  events(last: 3, topics: [$eventTopic], filter: "Leeched") {  
```

services/leecher/leecher/listener.py
```diff
@@ -28,12 +28,16 @@
 import shotgun_api3
 
 # TODO: remove hash in future since it is only used as backward compatibility
+# filters on only real events, starting with `Leeched`, not on `Recreated`
```
iLLiCiTiT, line 31: Either I don't understand the reasoning, or I don't understand the implementation. If we want to filter to specific topics, why we don't specify that in the `topics` filter?

iLLiCiTiT: BTW I didn't find `filter` in graphql, that is some generic graphql filtering option?

services/processor/processor/processor.py
```diff
@@ -244,8 +246,47 @@ def start_processing(self):
                             self,
                             payload,
                         )
-
+                    except MissingParentError:
```
iLLiCiTiT, line 249: Why we can't recreate parents instead of retriggering the event? Processor should not trigger events, it will add the event to the end of queue which might lead to huge issues as the task might change meanwhile.

services/leecher/leecher/listener.py
kalisp: There is a comment with 'wtf' for this, "Leeched", "leeched" will not work as a filter value, it will return empty list, even if there are events there.

iLLiCiTiT:
I am really against re-submitting leeched events in processor, it should not touch leeched events at all. At the first time of processing the event there might be other events changing or removing the same entity, so moving the event to the end of queue is a huge risk. There might be a real reason why parent was not synced, and re-submitting of the event can cause infinite loop of the event. It should be handled at the moment when we find out the parents are missing.

There also might be other services listening to leeched events, and re-submitting the events can cause their failure.

I'm not responsible for this addon, but for me this is no way.

services/leecher/leecher/listener.py
kalisp: I removed this weirdness and replaced by specific subtopic.

services/shotgrid_common/ayon_shotgrid_hub/__init__.py
```diff
@@ -306,6 +306,17 @@ def react_to_shotgrid_event(self, sg_event_meta):
...
+        # task revived only sends attribute_change event, before attribute
+        # change of parent assent AND finally entity_revival of Asset event
+        if (sg_event_meta["type"] == "attribute_change"
+            and sg_event_meta["attribute_name"] == "retirement_date"):
+            if sg_event_meta["entity_type"].lower() == "task":
+                self.log.info("changed to entity_revival")
+                sg_event_meta["type"] = "entity_revival"
+            else:
+                # do nothing for update retirement_date on non existing Asset
+                return
```
iLLiCiTiT, line 318:
```suggestion
            if sg_event_meta["entity_type"].lower() != "task":
                # do nothing for update retirement_date on non existing Asset
                return
            self.log.info("changed to entity_revival")
            sg_event_meta["type"] = "entity_revival"
```

services/leecher/leecher/listener.py
```diff
@@ -29,7 +29,7 @@
 
 # TODO: remove hash in future since it is only used as backward compatibility
 LAST_EVENT_QUERY = """query LastShotgridEvent($eventTopic: String!) {
-  events(last: 20, topics: [$eventTopic]) {
+  events(last: 20, topics: [$eventTopic]) {
```
iLLiCiTiT, line 32:
```suggestion
  events(last: 20, topics: [$eventTopic]) { 
```

iLLiCiTiT:
@MilaKudr could you re-test this please?

services/shotgrid_common/ayon_shotgrid_hub/update_from_shotgrid.py
```diff
@@ -172,8 +173,9 @@ def create_ay_entity_from_sg_event(
     if not ay_parent_entity:
         # This really should be an edge  ase, since any parent event would
         # happen before this... but hey
-        raise ValueError(
-            "Parent does not exist in Ayon, try doing a Project Sync.")
+        raise MissingParentError(
+            "Parent does not exist in Ayon, this event will be retried"
+            "after while. Hopefully parent will be already created.")
```
BigRoy, line 178: Missing space + grammar
```suggestion
            "Parent does not exist in Ayon, this event will be retried"
            " after a while. Hopefully parent will be already created.")
```

services/shotgrid_common/ayon_shotgrid_hub/update_from_shotgrid.py
```diff
@@ -284,7 +286,7 @@ def update_ayon_entity_from_sg_event(
     )
 
     if not ay_entity:
-        raise ValueError("Unable to update a non existing entity.")
+        raise MissingParentError("Unable to update a non existing entity.")
```
BigRoy, line 289: This may clarify where the entity is missing perhaps?
```suggestion
        raise MissingParentError("Unable to update a non existing AYON entity.")
```

services/shotgrid_common/ayon_shotgrid_hub/__init__.py
```diff
@@ -306,6 +306,16 @@ def react_to_shotgrid_event(self, sg_event_meta):
                 f"Ignoring event, AYON project {self.project_name} not found.")
             return
 
+        # task revived only sends attribute_change event, before attribute
+        # change of parent assent AND finally entity_revival of Asset event
```
BigRoy, line 310:
```suggestion
        # change of parent asset AND finally entity_revival of Asset event
```

services/processor/processor/processor.py
```diff
@@ -244,8 +247,46 @@ def start_processing(self):
...
+                        failed = True
+                        ayon_api.update_event(
+                            event["id"],
+                            status="failed",
+                            description=(
+                                "An error ocurred while processing"
+                                f"{event_id_text}, will be retried"
```
BigRoy, line 257: Missing space
```suggestion
                                "An error ocurred while processing "
                                f"{event_id_text}, will be retried"
```

BigRoy (commented):
Cosmetics

MilaKudr:
It is not working for me at the moment, after deleting the task in Ayon I get error messages like:
"2024-09-10 17:18:18.531 | update_from_ayon | remove_sg_entity_from_ayon_event | WARNING: Entity 'a82667b96f8711ef8ccdccd9acc131df' does not have a path to remove from Shotgrid.

kalisp:
I tried to rework logic without creating new event just by creating missing parents in the loop. Not most elegant code there though :/.

services/shotgrid_common/utils.py
```diff
@@ -1097,3 +1099,72 @@ def update_ay_entity_custom_attributes(
...
+
+    Folder Creation:
+        https://github.com/ynput/ayon-python-api/blob/30d702618b58676c3708f09f131a0974a92e1002/ayon_api/entity_hub.py#L254
+
+
+    Args:
+        entity_hub (ayon_api.EntityHub): The project's entity hub.
+        parent_entity: AYON parent entity.
+        sg_ay_dict (dict): AYON ShotGrid entity to create.
+    """
```
BigRoy, line 1123: Can we get the return type documented?

services/shotgrid_common/ayon_shotgrid_hub/__init__.py
```diff
@@ -306,6 +306,16 @@ def react_to_shotgrid_event(self, sg_event_meta):
...
             return
 
+        # task revived only sends attribute_change event, before attribute
+        # change of parent asset AND finally entity_revival of Asset event
+        if (sg_event_meta["type"] == "attribute_change"
+            and sg_event_meta["attribute_name"] == "retirement_date"):
+            if sg_event_meta["entity_type"].lower() != "task":
+                # do nothing for update retirement_date on non existing Asset
+                return
+            self.log.info("changed to entity_revival")
```
BigRoy, line 316: Can we make this log a bit clearer? This will be hard to grasp out of context if I'd see that log anywhere.

services/shotgrid_common/ayon_shotgrid_hub/__init__.py
```diff
@@ -306,6 +306,16 @@ def react_to_shotgrid_event(self, sg_event_meta):
                 f"Ignoring event, AYON project {self.project_name} not found.")
             return
 
+        # task revived only sends attribute_change event, before attribute
+        # change of parent asset AND finally entity_revival of Asset event
+        if (sg_event_meta["type"] == "attribute_change"
+            and sg_event_meta["attribute_name"] == "retirement_date"):
+            if sg_event_meta["entity_type"].lower() != "task":
```
BigRoy, line 313: The `retirement_date` change event I assume is ALSO triggered if you were to retire a task instead of reviving it.

Should we check the value of the retirement date being set?

Looking at [this](https://community.shotgridsoftware.com/t/retirement-dates/10545) my assumption is that the retirement date gets cleared out maybe when reviving? Should we hence also check the value whether it's really retiring it or not? Or do we want to handle the retiring in the same manner?

Or, do we even _care_ about processing the retirement date attribute events? Or are we actually syncing that shotgrid data too?

services/shotgrid_common/ayon_shotgrid_hub/update_from_ayon.py
```diff
@@ -253,20 +253,14 @@ def remove_sg_entity_from_ayon_event(
...
-    if not ay_entity_path:
-        log.warning(
-            f"Entity '{ay_id}' does not have a path to remove from Shotgrid."
-        )
-        return
-
     sg_id = ayon_event["payload"]["entityData"]["attrib"].get("shotgridId")
 
     if not sg_id:
+        ay_entity_name = ayon_event["payload"]["entityData"]["name"]
```
BigRoy, line 261: Why is this now 'name' instead of 'path'? I assume a path would be way clearer for the logs than just the name?

kalisp: Fixed

services/shotgrid_common/ayon_shotgrid_hub/__init__.py
kalisp: Updated logic a bit.

MilaKudr (approved):
It works for assets and tasks

PR #123, opened 2024-08-27
Integrate version: Support older python versions

tweak-wtf:
## Changelog Description
Changes the use of `|=` for updating dicts with `update` function in order to support python versions `<3.9`.

## Additional info
This was raised on discord --> https://discord.com/channels/517362899170230292/1248298172816822424/1277951309223428168

### DIFF
--- a/services/shotgrid_common/ayon_shotgrid_hub/match_ayon_hierarchy_in_shotgrid.py
+++ b/services/shotgrid_common/ayon_shotgrid_hub/match_ayon_hierarchy_in_shotgrid.py
@@ -241,13 +241,11 @@ def match_ayon_hierarchy_in_shotgrid(
         CUST_FIELD_CODE_ID: entity_hub.project_name,
         CUST_FIELD_CODE_SYNC: ay_project_sync_status
     }
-    data_to_update.update(
-        get_sg_custom_attributes_data(
-            sg_session,
-            entity_hub.project_entity.attribs.to_dict(),
-            "Project",
-            custom_attribs_map,
-        )
+    data_to_update |= get_sg_custom_attributes_data(
+        sg_session,
+        entity_hub.project_entity.attribs.to_dict(),
+        "Project",
+        custom_attribs_map,
     )
     sg_session.update(
         "Project",
@@ -388,13 +386,11 @@ def _create_new_entity(
             data[sg_parent_field] = sg_parent_entity
 
     # Fill up data with any extra attributes from Ayon we want to sync to SG
-    data.update(
-        get_sg_custom_attributes_data(
-            sg_session,
-            ay_entity.attribs.to_dict(),
-            sg_type,
-            custom_attribs_map
-        )
+    data |= get_sg_custom_attributes_data(
+        sg_session,
+        ay_entity.attribs.to_dict(),
+        sg_type,
+        custom_attribs_map
     )
 
     try:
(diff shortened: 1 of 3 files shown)


services/shotgrid_common/ayon_shotgrid_hub/match_ayon_hierarchy_in_shotgrid.py
```diff
@@ -241,11 +241,13 @@ def match_ayon_hierarchy_in_shotgrid(
...
-        entity_hub.project_entity.attribs.to_dict(),
-        "Project",
-        custom_attribs_map,
+    data_to_update.update(
+        get_sg_custom_attributes_data(
+            sg_session,
+            entity_hub.project_entity.attribs.to_dict(),
+            "Project",
+            custom_attribs_map,
+        )
```
BigRoy, line 250: I suspect the services wouldn't ever run client side - and hence would always run in the newer Python version on the server or however this service runs. So it _may be_ the case that we don't care about this file's changes.

@iLLiCiTiT @martastain for services, what Python version should we support? Client side code never touches this. Right?

tweak-wtf: whoops my bad.... afaik services have a hard dependency on python 3.9

tweak-wtf:
btw, i'll test this in-house in the evening.
i'll need to hotfix this in our studio but for that i have to fiddle a bit `package.py` because i don't want to build new images when only changing up client code.
so i gotta restart the services in a different addon version which means short downtime

services/shotgrid_common/ayon_shotgrid_hub/match_ayon_hierarchy_in_shotgrid.py
BigRoy: For context:

If we need to support Py3.10 or less then we should also refactor the match casing in Shotgrid addon, e.g. here: https://github.com/ynput/ayon-shotgrid/blob/7051f63c0980fe5a7e15631bb26c7be8a6be9448/services/shotgrid_common/ayon_shotgrid_hub/__init__.py#L227-L228

It's introduce in Py3.10: https://docs.python.org/3/whatsnew/3.10.html#pep-634-structural-pattern-matching

BigRoy: Also, if this needs no changing - then you'd only be deploying fixes client side. Not sure if that changes your "hotfix" case @tweak-wtf  ;)

tweak-wtf: just checked ayon-shotgrid's services python version dependency and they are set to ">=3.10,<4.0" so imho we don't need to touch the match casing.

referencing these:
https://github.com/ynput/ayon-shotgrid/blob/7051f63c0980fe5a7e15631bb26c7be8a6be9448/services/processor/pyproject.toml#L9
https://github.com/ynput/ayon-shotgrid/blob/7051f63c0980fe5a7e15631bb26c7be8a6be9448/services/leecher/pyproject.toml#L9
https://github.com/ynput/ayon-shotgrid/blob/7051f63c0980fe5a7e15631bb26c7be8a6be9448/services/transmitter/pyproject.toml#L9

tweak-wtf: > Also, if this needs no changing - then you'd only be deploying fixes client side. Not sure if that changes your "hotfix" case @tweak-wtf ;)

yeah unfortunately not quite. since i need to change code i also need to bump the addon version. i usually do something like `0.4.4-bepic.1` to indicate it's a "forked" version. i don't want to replace the `0.4.4` code on the server for consistency.

so because i'll have a new version for the SG addon but my services are still listening to `0.4.4` i assume they'll error out since they can't find the server settings in production bundle for this version.

tweak-wtf: so i'll revert the changes outside the client code area

iLLiCiTiT (approved):

Issue #126, opened 2024-09-03
AY-6666_Shotgrid: sequence and shot folder grouping


ynbot:
### Please describe the enhancement you have in mind and explain what the current shortcomings are?

For the Assets, the Flow->Ayon sync creates them into folders based on the Asset type ("sg_asset_type").
Which is a good thing and leads to a hierarchy in ayon like that:
<project_name>/<asset_type>/<asset_name>

But for the sequences (and I guess it will be the same for the shots at some point), Ayon creates them directly into the root, which leads to:
<project_name>/<sequence_name>

And we want:
<project_name>/sequence/<sequence_name>
And for the shots we'll want:
<project_name>/shot/<shot_name>

Is there a way to do that by setting things into Ayon ?

IMOO, the default behaviour in the code should be:
Take the entity's Type in Flow (Asset, Sequence, Shot...) and use it as main folder
For the Assets, use the sg_asset_type value with/instead (leading to <project_name>/asset/<asset_type>/<asset_name>, or <project_name>/<asset_type>/<asset_name> like it does now).

### Additional context:
link to discussion on [Discord]( )
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-6666](https://app.clickup.com/t/86c063myk)_

BigRoy:
Also came up on Community Discord [here](https://discord.com/channels/517362899170230292/1280666097976213545) - where it was mentioned that this may already be possible with a slight configuration tweak on Shotgrid side itself.

fabiaserra:
Sorry I misread this issue yesterday on my phone thinking that this referred to a common config issue where your Shotgrid projects `Tracking settings` don't have the right parent/child relationships set up and thus the entities appear flat on the hierarchy:
<img width="856" alt="Screenshot 2024-09-04 at 10 17 19 AM" src="https://github.com/user-attachments/assets/0214312f-8668-45b3-8040-c7dd7d4e1f2d">

Reading this now in the morning instead (with a coffee in my hands) I'm realizing this issue is indeed related to what @BigRoy linked in Discord at https://discord.com/channels/517362899170230292/1280666097976213545

However, this is requesting an even more granular control on the hierarchy and I'm not sure I understand the use case you are suggesting.

[...]

PR #129, opened 2024-09-10
Sequence and shot folder grouping

kalisp:
## Changelog Description
Synchronized `sequences` and `shots` were stored directly in root of the AYON project previously. This PR adds configurable option to set main `assets`, `sequences` and `shots` subdirectories.

image

## Additional info
Previously `seq0` would be directly next to `environment`, without wrapping `sequences` folder

There might be discussion needed for current implementation of `shots` without sequences in SG (if it makes actually sense in production). I implemented it according to requirements, but it might be just misunderstanding. 

! For running projects `ayon+settings://shotgrid/compatibility_settings/folder_locations` should be purged to keep original implementation. (Or maybe changed defaults OR changed initialization of this addon???)

Implemented 2 workflows:
image

This will reparent all SG `Assets` under `assets` category, all Sequences and their Shot children under `sequences`.
Resulting in:
image

##############################
Second use case 
image

Will separate SG `Assets` into AYON `assets`, SG `Sequences` into `sequences`, `Shots` into `/shots/new_shots`
Resulting in:
image

## Testing notes:

1. check `ayon+settings://shotgrid/compatibility_settings/folder_parenting` if it matches your requirements
2. synchronize from Shotgrid to AYON
3. `shots` folder is for shots without sequences (which is possible in SG)

### DIFF
--- a/server/settings/main.py
+++ b/server/settings/main.py
@@ -140,7 +140,7 @@ class FolderReparentingParentsModel(BaseSettingsModel):
         description=(
             "Name of the parent folder in AYON. Anatomy presets can be used."
             "`sg_` prefix can be used to refer to ShotGrid entities. Example: "
-            "`{sg_asset[type]}` will be replaced with the ShotGrid Asset Type."
+            "`{shotgrid_type}` will be replaced with the ShotGrid Asset Type."
         ),
     )
 
--- a/server/settings/main.py
+++ b/server/settings/main.py
@@ -113,6 +113,14 @@ class AttributesMappingModel(BaseSettingsModel):
     )
 
 
+class FolderLocationsModel(BaseSettingsModel):
+    """AYON folders to store separate of SG types"""
+    asset_folder: str = SettingsField(title="Assets", default="assets")
+    sequence_folder: str = SettingsField(title="Sequences",
+                                         default="sequences")
+    shot_folder: str = SettingsField(title="Shots", default="shots")
+
+
 class ShotgridCompatibilitySettings(BaseSettingsModel):
     """ Settings to define relationships between ShotGrid and AYON.
     """
@@ -133,6 +141,15 @@ class ShotgridCompatibilitySettings(BaseSettingsModel):
             "AYON attributes <> ShotGrid fields (without 'sg_' prefix!) "
             "mapping. Empty ones will be ignored. Scope is the list of "
             "ShotGrid entities that the mapping applies to. Disable any."
+        ),
+    )
+
+    folder_locations: FolderLocationsModel = SettingsField(
+        title="Folder locations",
+        default_factory=FolderLocationsModel,
+        description=(
+            "Locations of AYON folders matching to SG types."
+            "Eg. where SG assets will be stored, where SG shots and sequences."
         ),
     )
 
@@ -218,13 +235,13 @@ class ShotgridSettings(BaseSettingsModel):
     shotgrid_project_code_field: str = SettingsField(
         default="code",
         title="ShotGrid Project Code field name",
+        disabled=True,
         description=(
             "In order to create AYON projects, we need a Project Code, you "
             "can specify here which field in the ShotGrid Project "
             "entity represents it."
         ),
-        example="sg_code",
-        scope=["studio"],
+        example="sg_code"
     )
     enable_shotgrid_local_storage: bool = SettingsField(
         default=True,
--- a/server/settings/main.py
+++ b/server/settings/main.py
@@ -218,13 +218,13 @@ class ShotgridSettings(BaseSettingsModel):
     shotgrid_project_code_field: str = SettingsField(
         default="code",
         title="ShotGrid Project Code field name",
-        disabled=True,
         description=(
             "In order to create AYON projects, we need a Project Code, you "
             "can specify here which field in the ShotGrid Project "
             "entity represents it."
         ),
-        example="sg_code"
+        example="sg_code",
+        scope=["studio"],
     )
     enable_shotgrid_local_storage: bool = SettingsField(
         default=True,
--- a/server/settings/main.py
+++ b/server/settings/main.py
@@ -235,13 +235,13 @@ class ShotgridSettings(BaseSettingsModel):
     shotgrid_project_code_field: str = SettingsField(
         default="code",
         title="ShotGrid Project Code field name",
-        disabled=True,
         description=(
             "In order to create AYON projects, we need a Project Code, you "
             "can specify here which field in the ShotGrid Project "
             "entity represents it."
         ),
-        example="sg_code"
+        example="sg_code",
+        scope=["studio"],
     )
     enable_shotgrid_local_storage: bool = SettingsField(
         default=True,
--- a/server/settings/main.py
+++ b/server/settings/main.py
@@ -1,6 +1,10 @@
 from ayon_server.entities.core.attrib import attribute_library
 from ayon_server.settings import BaseSettingsModel, SettingsField
-from ayon_server.settings.enum import secrets_enum, anatomy_presets_enum
+from ayon_server.settings.enum import (
+    secrets_enum,
+    anatomy_presets_enum,
+    folder_types_enum
+)
 
 
 def default_shotgrid_entities():
@@ -27,6 +31,16 @@ def default_shotgrid_enabled_entities():
         "Asset",
         "Task",
         "Version",
+    ]
+
+
+def default_shotgrid_reparenting_entities():
+    """The entity types in ShotGrid that are enabled by default in AYON."""
+    return [
+        "Episode",
+        "Sequence",
+        "Shot",
+        "Asset",
     ]
 
 
@@ -113,13 +127,91 @@ class AttributesMappingModel(BaseSettingsModel):
     )
 
 
-class FolderLocationsModel(BaseSettingsModel):
-    """AYON folders to store separate of SG types"""
-    asset_folder: str = SettingsField(title="Assets", default="assets")
-    sequence_folder: str = SettingsField(title="Sequences",
-                                         default="sequences")
-    shot_folder: str = SettingsField(title="Shots", default="shots")
+class FolderReparentingParentsModel(BaseSettingsModel):
+    folder_type: str = SettingsField(
+        "asset",
+        title="Parent Ayon Folder Type",
+        enum_resolver=folder_types_enum,
+        description="Type of the parent folder in AYON",
+    )
+    folder_name: str = SettingsField(
+        "assets",
+        title="Parent Ayon Folder Name",
+        description=(
+            "Name of the parent folder in AYON. Anatomy presets can be used."
+            "`sg_` prefix can be used to refer to ShotGrid entities. Example: "
+            "`{sg_asset[type]}` will be replaced with the ShotGrid Asset Type."
+        ),
+    )
 
+
+class FolderReparentingPresetsModel(BaseSettingsModel):
+
+    filter_by_sg_entity_type: str = SettingsField(
+        "Asset",
+        title="Filter by ShotGrid Entity Type",
+        enum_resolver=default_shotgrid_reparenting_entities,
+        description=("Type of the ShotGrid entity to filter preset on."),
+    )
+    parents: list[FolderReparentingParentsModel] = SettingsField(
+        title="Parents",
+        default_factory=list,
+        description=(
+            "List of parent folders. If empty default behavior will be used. "
+            "The order of the parents from top to bottom is important."
+            "Within 'Root relocation' type the first parent will be "
+            "the root folder."
+        ),
+    )
+
+
+class FolderReparentingRelocateModel(BaseSettingsModel):
+    """Re-parent folders with Root relocation"""
+    enabled: bool = SettingsField(
+        False,
+        title="Enabled",
+        description="Enable or disable the re-parenting",
+    )
+
+    presets: list[FolderReparentingPresetsModel] = SettingsField(
+        title="Presets",
+        default_factory=list,
+        description=(
+            "List of presets for re-parenting. "
+            "If empty default behavior will be used."
+        ),
+    )
+
+
+class FolderReparentingTypeGroupingModel(BaseSettingsModel):
+    """Re-parent folders with Type grouping"""
+    enabled: bool = SettingsField(
+        False,
+        title="Enabled",
+        description="Enable or disable the re-parenting",
+    )
+
+    presets: list[FolderReparentingPresetsModel] = SettingsField(
+        title="Presets",
+        default_factory=list,
+        description=(
+            "List of presets for re-parenting. "
+            "If empty default behavior will be used."
+        ),
+    )
+
+
+class FolderReparentingModel(BaseSettingsModel):
+
+    root_relocate: FolderReparentingRelocateModel = SettingsField(
+        default_factory=FolderReparentingRelocateModel,
+        title="Root relocation",
+    )
+
+    type_grouping: FolderReparentingTypeGroupingModel = SettingsField(
+        default_factory=FolderReparentingTypeGroupingModel,
+        title="Type grouping",
+    )
 
 class ShotgridCompatibilitySettings(BaseSettingsModel):
     """ Settings to define relationships between ShotGrid and AYON.
@@ -144,13 +236,10 @@ class ShotgridCompatibilitySettings(BaseSettingsModel):
         ),
     )
 
-    folder_locations: FolderLocationsModel = SettingsField(
-        title="Folder locations",
-        default_factory=FolderLocationsModel,
-        description=(
-            "Locations of AYON folders matching to SG types."
-            "Eg. where SG assets will be stored, where SG shots and sequences."
-        ),
+    folder_parenting: FolderReparentingModel = SettingsField(
+        title="Folder re-parenting",
+        default_factory=FolderReparentingModel,
+        description=("Parent folders for AYON folders matching to SG types."),
     )
 
 
--- a/services/shotgrid_common/ayon_shotgrid_hub/match_shotgrid_hierarchy_in_ayon.py
+++ b/services/shotgrid_common/ayon_shotgrid_hub/match_shotgrid_hierarchy_in_ayon.py
@@ -109,23 +109,24 @@ def match_shotgrid_hierarchy_in_ayon(
 
         # If we couldn't find it we create it.
         if ay_entity is None:
-            if sg_ay_dict["attribs"].get(SHOTGRID_TYPE_ATTRIB) == "AssetCategory":  # noqa
+            parent_is_project = isinstance(ay_parent_entity, ProjectEntity)
+
+            shotgrid_type = sg_ay_dict["attribs"].get(SHOTGRID_TYPE_ATTRIB)
+            if shotgrid_type == "AssetCategory":
                 ay_entity = get_asset_category(
                     entity_hub,
                     ay_parent_entity,
                     sg_ay_dict
                 )
 
-            if (sg_ay_dict["attribs"].get(SHOTGRID_TYPE_ATTRIB) == "Sequence"
-                    and not ay_parent_entity):
+            if shotgrid_type == "Sequence" and parent_is_project:
                 ay_parent_entity = get_sequence_category(
                     entity_hub,
                     ay_parent_entity,
                     sg_ay_dict
                 )
 
-            if (sg_ay_dict["attribs"].get(SHOTGRID_TYPE_ATTRIB) == "Shot"
-                    and not ay_parent_entity):
+            if shotgrid_type == "Shot" and parent_is_project:
                 ay_parent_entity = get_shot_category(
                     entity_hub,
                     ay_parent_entity,
--- a/services/shotgrid_common/ayon_shotgrid_hub/match_shotgrid_hierarchy_in_ayon.py
+++ b/services/shotgrid_common/ayon_shotgrid_hub/match_shotgrid_hierarchy_in_ayon.py
@@ -189,7 +189,7 @@ def match_shotgrid_hierarchy_in_ayon(
                 CUST_FIELD_CODE_ID: ay_entity.id,
                 CUST_FIELD_CODE_SYNC: sg_entity_sync_status
             }
-            # Update Shotgrid entity with Ayon ID and sync status
+            # Update Shotgrid entity with AYON ID and sync status
             sg_session.update(
                 sg_ay_dict["attribs"][SHOTGRID_TYPE_ATTRIB],
                 sg_entity_id,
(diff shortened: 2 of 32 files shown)


fabiaserra:
I personally don't think this feature makes much sense. I imagine this comes after this request https://github.com/ynput/ayon-shotgrid/issues/126 ? As I commented there, I think it makes sense to group all asset entities under a folder called `assets` and all shot based entities under one called `shots` (or whatever) but splitting each entity into its own category beats the purpose in my opinion. Are shots parented to sequences going under the root `sequences` folder?

server/settings/main.py
```diff
@@ -218,13 +235,13 @@ class ShotgridSettings(BaseSettingsModel):
     shotgrid_project_code_field: str = SettingsField(
         default="code",
         title="ShotGrid Project Code field name",
+        disabled=True,
```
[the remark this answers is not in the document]
kalisp: Fffff...unwanted, reverted. Thanks for noticing.

kalisp:
`Are shots parented to sequences going under the root sequences folder` they would now (but they shouldn't according to requirements, `sequences` should contain only sequences)

I am open to ideas how to configure it to be more universal as this is sponsored  issue, but I agree it might not apply to all use cases, same as yours doesn't.

fabiaserra:
> `Are shots parented to sequences going under the root sequences folder` they would now (but they shouldn't according to requirements, `sequences` should contain only sequences)

So where are shots parented to sequences go? Or a sequence child of an episode and that episode having sequences and then shots? Could you put some examples of the different cases?

BigRoy:
I'm so clueless about these conversations - doesn't shotgrid just have hierarchies itself? And shouldn't we just adhere to that hierarchy 1:1? Or what makes this so complex?

fabiaserra:
> I'm so clueless about these conversations - doesn't shotgrid just have hierarchies itself? And shouldn't we just adhere to that hierarchy 1:1? Or what makes this so complex?

Yeah I'm a bit confused too tbh, I thought the request was to have it just on disk (which wouldn't require to create the entities in the server and it's just a template thing) and just as a means to encapsulate ALL of the shots (with its respective hierarchy) within. On the other hand, if they want those categories to show up on the UIs that's something we can do in the frontend agnostic of shotgrid as the folder types exist, it's basically just an extension to the existing icons?

tweak-wtf:
@antirotor asked me to chime in to explain my usecase as i'm also interested in this feature and we've talked about it last week.
lemme share my thoughts on this...

SG doesn't have the ability to structure entities into generic folders as AYON.
Using SG's project tracking features allows you to create hierarchies like `Episode->Sequence->Shot` however it's currently not possible to place these SG hierarchies in a custom AYON folder.
Syncing AYON entities to SG is easy since we just need `folder_name` and `folder_type` of AYON and create the SG entities respectively as SG takes care of the linking.

so I believe this issue boils down to being able to use custom AYON folders as roots for SG hierarchies.

assume an SG project tracking setting for assets as `AssetType->Asset` and the following AYON hierarchy for assets:
```
├── build
│   └── assets
│       └── Prop
│       ├── Character
│       └── Environment
``` 

I want to have my SG-created assets to show up in `/build/assets/{asset[type]}/` in AYON.

### Suggestion
We get SG's project tracking settings for enabled entity types and expose them as part of a new settings field in the compatibility settings. Each SG tracking setting gets a string field representing the AYON folder hierarchy to follow.
E.g.:

SG Asset: AssetType->Asset    -->     AYON hierarchy to follow: /build/assets/{asset[type]}/

If a user doesn't want to use custom AYON hierarchies and basically start with a blank AYON project populated from SG they'd leave all lineedits empty and SG's project tracking settings will be followed (even if flat? error otherwise?).

I think this way we could support any AYON hierarchy with any SG project tracking.

fabiaserra:
> Using SG's project tracking features allows you to create hierarchies like Episode->Sequence->Shot however it's currently not possible to place these SG hierarchies in a custom AYON folder.

that's not entirely true? I can change the project anatomy template and place all of those entities under any custom folder that I want? What it's true is that those wouldn't be on different individual custom folders but that's why I raised my doubts, on which case would you want that? Why would you want the shots children of a specific sequence under a different custom folder? And if that's not what you want and you basically just want like what we have in our fork, do you just want two separate custom folders for the `assets` type entities (build) and the `shots` (episode, sequence, shot...)? If the latter, the solution is much simpler than what's being suggested here.

> Syncing AYON entities to SG is easy since we just need folder_name and folder_type of AYON and create the SG entities respectively as SG takes care of the linking.

Sorry but that's not entirely true either, unless you set the parent/child relationships in SG those entities will be childless

> I want to have my SG-created assets to show up in /build/assets/{asset[type]}/ in AYON.

This kind of points me to believe what you want is what I was referring to simply separating the `assets` entities and the `shots`-based entities and so if that's the case, I don't think we should be doing any changes to `ayon-shotgrid` and it should all be `ayon-core`, basically allowing for more granular control on the template syntax to express the different types of folders. In our fork, as I commented here https://github.com/ynput/ayon-shotgrid/issues/126#issuecomment-2329235843, I added support for a new key on the template called `{assets_or_shots}` that basically contains the string `shots` or `assets` depending on what type of entity it is. Expanding on that... it would be pretty simple to make that token contain `/build/assets` or `/shots` instead and those changes would only require a few extra lines of code and avoid adding extra complexity to this codebase AND it would work for all the other project management addons (ftrack, kitsu...) and not be specific to this

tweak-wtf:
my specific use case is to create SG Assets under `/build/assets` in AYON following the SG tracking settings to result in `/build/assets/{asset[type]}/{asset[name]}`.
i don't want to break the integrity of SG tracking settings. rather use them under a new root.
no SG shots linked to an SG Sequence broken apart into different folders in AYON or so.

it's more about how to treat asset types that are parentless in terms of entity tracking settings, like Sequence (by default?) in SG.
To do that i'm suggesting to merge an implicit SG tracking settings path and an explicit AYON path together, optionally.
e.g. tracking setting `Sequence->Shot` for an SG Shot `sh010` in sequence `seq01` with AYON folder `/sequences` would build `/sequences/seq01/sh010`. and if a user creates an SG Shot with no SG Sequence? yikes, but we could place it in the next available root in `/sequences/<NONE>/sh010` which would be `/sequences`.

in #126 we unfortunately don't have the SG tracking settings to check which would help and tbh i don't quite get that use case either. might be all flat tracking settings in SG, might be breaking `Sequence->Shot` tracking. 
for all everything flat i could imagine having a bunch of parentless shots being assembled by different sequences/edits but i'm assuming.

fabiaserra:
> my specific use case is to create SG Assets under `/build/assets` in AYON following the SG tracking settings to result in `/build/assets/{asset[type]}/{asset[name]}`. i don't want to break the integrity of SG tracking settings. rather use them under a new root. no SG shots linked to an SG Sequence broken apart into different folders in AYON or so.
> 
> it's more about how to treat asset types that are parentless in terms of entity tracking settings, like Sequence (by default?) in SG. To do that i'm suggesting to merge an implicit SG tracking settings path and an explicit AYON path together, optionally. e.g. tracking setting `Sequence->Shot` for an SG Shot `sh010` in sequence `seq01` with AYON folder `/sequences` would build `/sequences/seq01/sh010`. and if a user creates an SG Shot with no SG Sequence? yikes, but we could place it in the next available root in `/sequences/<NONE>/sh010` which would be `/sequences`.
> 
> in #126 we unfortunately don't have the SG tracking settings to check which would help and tbh i don't quite get that use case either. might be all flat tracking settings in SG, might be breaking `Sequence->Shot` tracking. for everything flat i could imagine having a bunch of parentless shots being assembled by different sequences/edits but i'm assuming.

I'm honestly still pretty confused as to what you are asking. Have you played much with the current `ayon-shotgrid`, the anatomy templates and tried to sync different entities from SG? If entities don't have a parent they are already going to a flat hierarchy as I would expect and the current implementation of `ayon-shotgrid` DOES use the project tracking settings as a means to know what's the hierarchy of the entities that you want to map in AYON.

For example, we have these parentless shots on all SG projects: `_edit_shot`, `_2d_shot` and `_3d_shot` and when we sync them into AYON they go under this location:
```
> ll /proj/dev_000/shots/
total 0
0 drwxrwsrwx 10 root ax_users  8 Jan 18  2024 ./
0 drwxrwsrwx 17 root ax_users 15 Sep 27 03:05 ../
0 drwxrwsrwx  6 root ax_users  4 Jan 24  2024 000_seq/
0 drwxrwsrwx  4 root ax_users  2 Jan 24  2024 _2d_shot/
0 drwxrwsrwx  4 root ax_users  2 Mar 13  2024 _3d_shot/
0 drwxrwsrwx  3 root ax_users  1 Jan  3  2024 _edit_shot/
```
And our `assets` are going under here:
```
> ll /proj/dev_000/assets/
total 0
0 drwxrwsrwx  6 root ax_users  4 Jul 26  2023 ./
0 drwxrwsrwx 17 root ax_users 15 Sep 27 03:05 ../
0 drwxrwsrwx  5 root ax_users  3 May 10 13:00 Prop/
0 drwxrwsrwx  2 root ax_users  0 May 10  2023 Characters/
0 drwxrwsrwx  2 root ax_users  0 May 10  2023 Locations/
```

It seems like the only change you want on top of what I already have is to have the `assets` under an extra nested folder called `build`, the rest would be the same? and as I suggested, adding that extra folder should be much simpler than what you are suggesting doing here unless I'm not understanding correctly

tweak-wtf:
true, i haven't used the anatomy presets since i thought it's only used when creating an AYON project from an existing SG one. i always start off projects in AYON with a predefined folder hierarchy and have the docker stack create the SG project.
when syncing SG entities to AYON I know they are following their tracking settings. but they do not respect existing AYON folder hierarchy. that's why i have to tell prods/coords that they need to create entities in AYON and not in SG. however, i would like to give them the freedom to create entities in AYON or SG, their choice.

i'm having some issues explaining my use case more than by saying: i want to be able to sync SG entities to a configurable root folder in AYON in order to support any custom AYON hierarchy respecting SG's tracking settings for that entity. 

however, if i get u right using the anatomy templates is able to create SG entities in AYON under a custom root folder.
so with the latest versions of `ayon-core` and `ayon-shotgrid`, which template settings of an anatomy preset do i need to change to achieve SG assets being synced to `/assets` (or `/build/assets`) then following the tracking settings `AssetType->Asset` just as u do?

always down for having a quick call btw.

fabiaserra:
> true, i haven't used the anatomy presets since i thought it's only used when creating an AYON project from an existing SG one. 

Not really, anatomy presets are agnostic from SG, they are just a means to set how the entities live on disk

> i always start off projects in AYON with a predefined folder hierarchy and have the docker stack create the SG project. when syncing SG entities to AYON I know they are following their tracking settings. but they do not respect existing AYON folder hierarchy. that's why i have to tell prods/coords that they need to create entities in AYON and not in SG. however, i would like to give them the freedom to create entities in AYON or SG, their choice.

Is this because you are creating an actual AYON folder entity named "build" for you to have that extra grouping for `/build/assets/...` as opposed to just having that `build` and `assets` folders ONLY on disk? In our server, the same hierarchy that we have in AYON is mapped exactly the same in SG and viceversa so we are able to create the entities anywhere and they will look the same and the creation of the `shots` and `assets` folders are ONLY operating system folders, not the "AYON Folder", we simply set these folders on the anatomy preset so this extra folder gets created on the project structure on disk

> i'm having some issues explaining my use case more than by saying: i want to be able to sync SG entities to a configurable root folder in AYON in order to support any custom AYON hierarchy respecting SG's tracking settings for that entity.

If I'm understanding your problem correctly, I think the issue is that you are trying to create these kind of typeless folders on AYON ("build", "assets", "shots"...) and that makes sense why it's an issue on the syncing between both as those extra folder hierarchy don't make much sense in Shotgrid. What's your use case for having those extra folders in AYON though and not simply on disk?

> however, if i get u right using the anatomy templates is able to create SG entities in AYON under a custom root folder. so with the latest versions of `ayon-core` and `ayon-shotgrid`, which template settings of an anatomy preset do i need to change to achieve SG assets being synced to `/assets` (or `/build/assets`) then following the tracking settings `AssetType->Asset` just as u do?

You are not able to do that with the current state of `ayon-core` and `ayon-shotgrid` but in our fork of `ayon-core`, we did this simple change on the template_data: https://github.com/fabiaserra/ayon-core/blob/release/alkemyx/client/ayon_core/pipeline/template_data.py#L107 that allows us to use this token on the anatomy `{assets_or_shots}` that basically contains the string `assets` or `shots` and when the entities get created/referenced on disk that path contains one or the other. Extending on that functionality, one could expose something similar on the settings so users could configure what the value of this token contains depending on the folder type, so you would be able to define that for "asset" folder types... this string contains `/build/assets` instead of just `assets`, and similarly, for `episodes`, `sequences` and `shots` they could contain `shots` or any other folder structure. But as I said, I think that should only be a matter of extending the functionality of the anatomy templates so you have more control on how the folders get created ON disk.

However, if I understood correctly, your problem is that these custom folders are not only on disk, they are actual extra "AYON Folders" that you want to have in your project. If that's the case, I think that could be something supported on the AYON server itself where you are able to create extra grouping folders and on the `ayon-shotgrid` you would be able to choose which ones to ignore so the hierarchy just bypasses them

tweak-wtf:
> If I'm understanding your problem correctly, I think the issue is that you are trying to create these kind of typeless folders on AYON ("build", "assets", "shots"...) and that makes sense why it's an issue on the syncing between both as those extra folder hierarchy don't make much sense in Shotgrid. What's your use case for having those extra folders in AYON though and not simply on disk?

yes exactly!
i don't really see them as typeless since afaik they do have a type, it's just `Folder` 
so basically every time i said folder i was referring to an "AYON Folder" of type `Folder`. oh jeez the lingo :D sorry if that didn't come across.
tbh, i just kinda like the resemblance to the os filesystem of AYON's browser. i can imagine doing more cool stuff with it than an os filesystem folder. i like to have AYON folders created based on some template to init an AYON project's hierarchy prefilled with some test shots, assets, tasks and such and have those synced back to SG.
that's what i meant with: SG doesn't have the ability to structure entities into generic folders as AYON.
i believe kitsu doesn't, aquarium should (it's node-based) and i don't know about the others.

> However, if I understood correctly, your problem is that these custom folders are not only on disk, they are actual extra "AYON Folders" that you want to have in your project. If that's the case, I think that could be something supported on the AYON server itself where you are able to create extra grouping folders and on the ayon-shotgrid you would be able to choose which ones to ignore so the hierarchy just bypasses them

yeah, extra actual typeless AYON folders "groups(?)" could be a thing!
OT: would also love to see product grouping on the AYON server's browser synced to a dcc desktop loader.

kalisp:
> However, if I understood correctly, your problem is that these custom folders are not only on disk, they are actual extra "AYON Folders" that you want to have in your project.

Yes, it started as client request where they didnt like that shots and sequences weren't grouped. 
State without this PR. 
chrome_Al2KHP1JC1
image

fabiaserra:
> yes exactly! i don't really see them as typeless since afaik they do have a type, it's just `Folder` so basically every time i said folder i was referring to an "AYON Folder" of type `Folder`. oh jeez the lingo :D sorry if that didn't come across.

Yeah... it's quite confusing and I wonder if the term Ynput chose actually is what drove you to end up doing this workflow hahaha I personally don't see the disk folder structure the same way as the entity hierarchy on AYON or SG and I would hope the icons are already kind of acting the same purpose as giving a "category" to the different entities without needing an extra folder, but I guess I could see someone preferring the extra categorization by adding extra folders.

> SG doesn't have the ability to structure entities into generic folders as AYON. i believe kitsu doesn't, aquarium should (it's node-based) and i don't know about the others.

I don't think that's an AYON only thing though, you could probably do the same in SG by creating custom entities that can be parented to other existing entities so they act like "AYON Folders", although as I said, I think that would complicate things quite a bit as all of a sudden a lot of the assumptions we can do in most pipelines working on features/episodic/animation are no longer that true and it certainly makes this `ayon-shotgrid` addon much more complex and error-prone (similar to having to do the same for `ayon-ftrack` and any other sync addons)

> yeah, extra actual typeless AYON folders "groups(?)" could be a thing! 

I think that's how the solution to this issue should be implemented TBH, supporting "dummy category" folders in AYON natively so they can be easily ignored on the different sync addons that don't support the same structure by design

> OT: would also love to see product grouping on the AYON server's browser synced to a dcc desktop loader.

Probably misunderstanding what you mean but product groups do show up on the loaders of AYON?

fabiaserra:
> Yes, it started as client request where they didnt like that shots and sequences weren't grouped. State without this PR. chrome_Al2KHP1JC1 image

I think this image is particularly misleading as that Shot doesn't have any episode or sequence as parent. If it did, the `sh010` wouldn't be sitting there at the roof of the project. You can always name your entities however you want and if for example the episodes were named `ep101` and the sequence `seq_tst`, that shot would be on a hierarchy of `ep101/seq_tst`, which is already pretty easy in my opinion to see the difference of hierarchy between the `assets` and the `shots`

kalisp:
Yet client would expect something like `episodes/ep101/seq_tst`.

fabiaserra:
> Yet client would expect something like `episodes/ep101/seq_tst`.

Is that already working with the work of this PR? Could you provide some production examples of the category folders created in a project? Can they have any number of nested category folders or just one?

dee-ynput:
> Could you provide some production examples of the category folders created in a project? Can they have any number of nested category folders or just one?

Hello @fabiaserra 
This PR is implementing our client request. If you're interested in production examples I invite you to deploy and test it, or wait for the release 👍
Thanks for your participation in this thread, it's always usefull to challenge implementation.
The request iteself is not up for debate though.

If some supporting users ask for the `{asset_or_shot}` token, we'll be more than happy to add it and it would make you way of dealing with hierarchy available too.
That is not our goal here though.

Cheers!

fabiaserra:
> Hello @fabiaserra This PR is implementing our client request. If you're interested in production examples I invite you to deploy and test it, or wait for the release 👍 

I'm interested because merging your releases can break other people's forks of `ayon-shotgrid`, like ours, and discussing about other options it's good for you and the rest of the community. Asking for a more clear explanation of what the PR tries to do is just the bare minimum of a controversial PR. If you are 100% confident about the implementation go ahead, we will do our own assessment when we try merge the next release (if we do)

> The request iteself is not up for debate though.

I wasn't debating the request, just the implementation of it and helping to find out a better solution

> Thanks for your participation in this thread, it's always usefull to challenge implementation.

You are welcome but honestly things like this just disencourage me in continuing to participate

services/shotgrid_common/ayon_shotgrid_hub/match_shotgrid_hierarchy_in_ayon.py
```diff
@@ -104,11 +107,31 @@ def match_shotgrid_hierarchy_in_ayon(
 
         # If we couldn't find it we create it.
         if ay_entity is None:
-            if sg_ay_dict["attribs"].get(SHOTGRID_TYPE_ATTRIB) == "AssetCategory":  # noqa
+            parent_is_project = isinstance(ay_parent_entity, ProjectEntity)
```
jakubjezek001, line 110:
```suggestion
            parent_is_project = ay_parent_entity.entity_type == "project"
```

jakubjezek001: `ProjectEntity` object does not exists.

jakubjezek001:
@kalisp it seems some uncommited changes are still on your machine and need to be pushed. 

My tests were showing some ProjectEntity rises (metnioned in code). And also here is output of the project sync. 
image

This is working only partly. Shots should also be in its own folder `Shots` so it is alligned with your image.

jakubjezek001 (changes requested):
Perhaps some some local changes needs to be pushed so this could be reviewed again.

services/shotgrid_common/ayon_shotgrid_hub/match_shotgrid_hierarchy_in_ayon.py
kalisp: Fixed.

tweak-wtf:
the suggested server setting look great. exactly what i had in mind. This also gives enough flexibility to account for entities other than `Shot` or `Asset`, e.g. `Lens`.

> It is important to mention that we need to have Shotgrid Entity type filtering for each preset, so we could add ultiple entity type related presets. Yes it does have a caviate and it is possible multiple usage of single entity type with different reparenting types. This would ofcourse could kill the synchronization so we would need to find a way to validate those presets.

good catch! also we'd have to check against SG tracking settings.
with these new AYON server settings a user might end up configuring different AYON parent folders for `Sequence` and `Shot` while using a SG tracking setting like `Sequence->Shot`. Then the service wouldn't know what to do as it would try to create the sequence in a different location than the shot effectively breaking the link. not hehe.
seems this validation could turn out to be somewhat tricky. maybe we could (for now) restrict the presets to be configured to `Shots` and `Assets` only?  maybe this could also be handled by simply writing good docs that warn a user about potentially faulty configurations, maybe giving examples and why they don't make sense.

at this point i just wanna note again that this is all about following an SG entities tracking setting in AYON under a specific parent folder, not about breaking those hierarchy links.

> I believe that this settings model should be able provide the universal functionality we would need for such feature. What do you thing @tweak-wtf, would this be fiesable for your production?

yes i completely agree with everything u wrote. love the server settings approach which is more than feasible, although validating erroneous user preset configuration might become a challenge.

jakubjezek001:
@tweak-wtf please look into #147 for some update on the settings.

MilaKudr (approved):
Both methods work fine for me. For some reason there is a one minute delay (waiting) between pressing the Sync button (SG -> AYON) since the Processor service begin to do something, lets see if this can be improved.

jakubjezek001 (changes requested):
```
2024-11-29 12:03:58.215 INFO: Event has been processed... setting to finished!
2024-11-29 12:03:58.971 ERROR: Unable to create Task <9f491990ae4111efb21b85431f85a5f6> in Shotgrid!
  File "C:\CODE\__YNPUT\ayon-shotgrid\services\shotgrid_common\ayon_shotgrid_hub\update_from_ayon.py", line 83, in create_sg_entity_from_ayon_event
    sg_entity = _create_sg_entity(
                ^^^^^^^^^^^^^^^^^^
  File "C:\CODE\__YNPUT\ayon-shotgrid\services\shotgrid_common\ayon_shotgrid_hub\update_from_ayon.py", line 340, in _create_sg_entity
    raise ValueError(
ValueError: Parent does not exist in Shotgrid!None <None>
```

This is happening only with Transmitter once a AY project is set to `shotgrid_push` activated for auto syncing based on AY changes. 

*****
Edit: 
So the another examination of the issue discovered following issue at this testing:
- Shots in Relocate Root workflow is synchronized but parent (in AY under a Sequence) is not correctly set. 
- The above issue was caused by wrong setting Folder type on shot (Shotgrid is not supporting Folder type)

jakubjezek001:
We will be sorting the parenting issue in consequent PR @kalisp please. But let's not do new release until the parenting is solved @m-u-r-p-h-y @tatiana-ynput .

jakubjezek001 (approved):

PR #131, opened 2024-09-13
Filling SG versions path_to_frames

ReeceMulley:
## Changelog Description
My studio requires uploaded versions' sg_path_to_frames field to be filled in with the EXRs to review / deliver. This checks all representations attached to the product, does basic filtering and takes the path from there. 

## Additional info
If a product has >1 compatible image representations (created through the oiiotool for example), the last representation's path is used. Last alphabetically, as far as I can tell. Perhaps a tag could be added on the Extract Review output definition to specify which representation name to use, ie "path_to_frames:exr2065". But I'm not sure if there's existing conventions for this, and we needed a hotfix

## Testing notes:

1. Publish a review to SG. I did a nuke render with stock Extract Review settings
2. Check "sg_path_to_frames" field against the representation's path in ayon

### DIFF
--- a/client/ayon_shotgrid/plugins/publish/integrate_shotgrid_version.py
+++ b/client/ayon_shotgrid/plugins/publish/integrate_shotgrid_version.py
@@ -43,66 +43,69 @@ class IntegrateShotgridVersion(pyblish.api.InstancePlugin):
         # find thumbnail path
         thumbnail_path = instance.data.get("thumbnailPath")
 
+        # set pathtoframes exclusion filters
+        frames_path_filters = ["thumb", "workfile"]
+
         found_reviewable = False
+        found_frames = False
+
         for representation in instance.data.get("representations", []):
             self.log.debug(pformat(representation))
 
-            if "shotgridreview" not in representation.get("tags", []):
+            # Skip unwanted representations
+            if any(filter_name for filter_name in frames_path_filters if filter_name in f".{representation['name']}"):
                 continue
 
             local_path = get_publish_repre_path(
                 instance, representation, False
             )
 
-            if f".{representation['ext']}" in VIDEO_EXTENSIONS:
-                found_reviewable = True
-                data_to_update["sg_path_to_movie"] = local_path
-                for representationSecondary in instance.data.get("representations", []):
-                    if f".{representationSecondary['ext']}" in IMAGE_EXTENSIONS and "thumbnail" not in f".{representationSecondary['name']}":
-                        frames_path = get_publish_repre_path(
-                            instance, representationSecondary, False
-                        )
-                        path_to_frames = re.sub(r"\.\d+\.", ".####.", frames_path)
-                        data_to_update["sg_path_to_frames"] = path_to_frames
-                        self.log.debug(f"setting path_to_frames to {path_to_frames}")
-                if (
-                    "slate" in instance.data["families"]
-                    and "slate-frame" in representation["tags"]
-                ):
-                    data_to_update["sg_movie_has_slate"] = True
+            # Skip temp locations
+            if local_path.startswith("/tmp"):
+                continue
 
-            elif f".{representation['ext']}" in IMAGE_EXTENSIONS:
-                found_reviewable = True
-                # Replace the frame number with '%04d'
-                path_to_frame = re.sub(r"\.\d+\.", ".%04d.", local_path)
+            # Replace the frame number with '####'
+            local_path = re.sub(r"\.\d+\.", ".####.", local_path)
 
-                data_to_update.update({
-                    "sg_path_to_movie": path_to_frame,
-                    "sg_path_to_frames": path_to_frame,
-                })
-                for representationSecondary in instance.data.get("representations", []):
-                    if f".{representationSecondary['ext']}" in IMAGE_EXTENSIONS:
-                        if "thumb" not in f".{representationSecondary['name']}" and "review" not in f".{representationSecondary['name']}":
-                            frames_path = get_publish_repre_path(
-                                instance, representationSecondary, False
-                            )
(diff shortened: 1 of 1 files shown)


fabiaserra:
the logic feels wrong as you are doing a second iteration of all the representations within the main loop of representations.

In our fork of `ayon-shotgrid` I'm skipping the `thumbnail` repre always as we never want that to be integrated on the version https://github.com/fabiaserra/ayon-shotgrid/blob/release/alkemyx/client/ayon_shotgrid/plugins/publish/integrate_shotgrid_version.py#L58 and also the latest image extension on the representations will be set to `sg_path_to_frames` without having to do another for loop.

ReeceMulley:
> the logic feels wrong as you are doing a second iteration of all the representations within the main loop of representations.
Hey! I've changed this up a bit - now the second iteration isn't happening. We've also added a reference to the representation tag "shotgridpathtoframes" - if a compatibles representation with this is found, these frames are used. 
Also, I've used #### for the frame number replacement - this is important for our use case but %04d may be preferred, idk

client/ayon_shotgrid/plugins/publish/integrate_shotgrid_version.py
```diff
@@ -43,38 +43,58 @@ def process(self, instance):
...
+
         found_reviewable = False
+        found_frames = False
+
         for representation in instance.data.get("representations", []):
             self.log.debug(pformat(representation))
 
-            if "shotgridreview" not in representation.get("tags", []):
+            # Skip unwanted representations
+            if any(filter_name for filter_name in frames_path_filters if filter_name in f".{representation['name']}"):
```
fabiaserra, line 56:
```suggestion
            if any(filter_name for filter_name in frames_path_filters if filter_name in representation['name']):
```

client/ayon_shotgrid/plugins/publish/integrate_shotgrid_version.py
```diff
@@ -43,38 +43,58 @@ def process(self, instance):
         # find thumbnail path
         thumbnail_path = instance.data.get("thumbnailPath")
 
+        # set pathtoframes exclusion filters
+        frames_path_filters = ["thumb", "workfile"]
```
fabiaserra, line 47: how about calling this variable something like `excluded_representations`?

client/ayon_shotgrid/plugins/publish/integrate_shotgrid_version.py
```diff
@@ -43,38 +43,58 @@ def process(self, instance):
...
+            # Skip temp locations
+            if local_path.startswith("/tmp"):
+                continue
 
-                if "slate" in instance.data["families"]:
-                    data_to_update["sg_frames_have_slate"] = True
+            # Replace the frame number with '####'
+            local_path = re.sub(r"\.\d+\.", ".####.", local_path)
+
+            if "shotgridreview" not in representation.get("tags", []) and not found_frames:
```
fabiaserra, line 70: can you explain the logic behind this?

client/ayon_shotgrid/plugins/publish/integrate_shotgrid_version.py
```diff
@@ -85,7 +105,7 @@ def process(self, instance):
             })
 
         # If there's no data to set/update, skip creation of SG version
-        if not found_reviewable:
+        if not found_reviewable or "workfile" in f"{instance.data['productName']}":
```
fabiaserra, line 108: why not just checking `if not data_to_update`? check our fork https://github.com/fabiaserra/ayon-shotgrid/blob/release/alkemyx/client/ayon_shotgrid/plugins/publish/integrate_shotgrid_version.py#L112

client/ayon_shotgrid/plugins/publish/integrate_shotgrid_version.py
```diff
@@ -43,38 +43,58 @@ def process(self, instance):
...
-                found_reviewable = True
-                # Replace the frame number with '%04d'
-                path_to_frame = re.sub(r"\.\d+\.", ".%04d.", local_path)
-
-                data_to_update.update({
-                    "sg_path_to_movie": path_to_frame,
-                    "sg_path_to_frames": path_to_frame,
-                })
+            # Skip temp locations
+            if local_path.startswith("/tmp"):
```
robin-ynput, line 64: I'm concerned this is too platform-specific, maybe we should check against anatomy root path instead ?

client/ayon_shotgrid/plugins/publish/integrate_shotgrid_version.py
robin-ynput: If I understand the code, I think this changes a bit the logic..

What used to be:
1. inspect whatever representation was found under `shotgridreview` tag 
2. fill up `sg_path_to_movie` and/or `sg_path_to_frames`/`sg_frames_have_slate` accordingly

becomes:
1. loop over all representations. As soon as a non-temp img-seq repr. is found, use it for `sg_path_to_frames`
2. potentially override it with `shotgridreview` tagged repr. unless it was tagged `shotgridpathtoframes`

What I'm unsure with is that there does not seem to be any guaranteed order between `shotgridpathtoframes` and `shotgridreview`. It'll be depending on which representation comes first.

Maybe that should be:
1. attempt to detect `shotgridpathtoframes` and use it if found
2. otherwise try to get it from `shotgridreview` 
3. last resort, fill up with any other img-sequence matching repr.
 (_Personally I'm doubtful about this one cause I feel like if a `shotgridpathtoframes` tag feature is available, we might as well enforce it instead of providing a random default)_

robin-ynput:
Hello there, it looks like this PR is stalled and not actively worked on anymore.
As this is a community feature request, I'm going to close it for now, but please feel free to re-open if someone is willing to pursue the work and answer the comments here.

PR #132, opened 2024-09-13
Ignoring Task entities with no step key

jakubjezek001:
# Change description
Some productions might be using tasks without SG task entity `step` key. This will change will not raise the missing key and therefore it will allow to sync tasks into project. New settings for defining default task type were added under Compatibility Settigs section. This way we can assign task type to those tasks which are missing pipeline step at SG entity. 

# Testing steps:
1. Make sure `ayon+settings://shotgrid/compatibility_settings/default_task_type` is set at your Project settings with existing Anatomy Task type. Ideally create new one and name it `Unknown`
2. Go to any of yours test project at SG and create new task under any Shot or Asset.
3. Make usre the autosync is set to both direction on SG and AY side.
4. Task shoud be created and the default task type is assigned to it. 
5. Remove the task on AY side and notice that after refreash it had been also retired on SG side.

### DIFF
--- a/services/shotgrid_common/utils.py
+++ b/services/shotgrid_common/utils.py
@@ -116,19 +116,23 @@ def _sg_to_ay_dict(
     if sg_entity["type"] == "Task":
         ay_entity_type = "task"
         if not sg_entity["step"]:
-            raise ValueError(
+            log.warning(
                 f"Task {sg_entity} has no Pipeline Step assigned."
             )
+            task_type = None
+        else:
+            task_type = sg_entity["step"]["name"]
 
         label = sg_entity["content"]
-        task_type = sg_entity["step"]["name"]
 
         if not label and not task_type:
             raise ValueError(f"Unable to parse Task {sg_entity}")
+
         if label:
             name = slugify_string(label)
-        else:
+        elif task_type:
             name = slugify_string(task_type)
+
 
     elif sg_entity["type"] == "Project":
         name = slugify_string(sg_entity[project_code_field])
--- a/services/shotgrid_common/utils.py
+++ b/services/shotgrid_common/utils.py
@@ -121,7 +121,7 @@ def _sg_to_ay_dict(
     if sg_entity["type"] == "Task":
         ay_entity_type = "task"
         if not sg_entity["step"]:
-            log.warning(
+            log.debug(
                 f"Task {sg_entity} has no Pipeline Step assigned. "
                 "Task type set from settings."
             )
--- a/services/shotgrid_common/utils.py
+++ b/services/shotgrid_common/utils.py
@@ -111,6 +111,7 @@ def _sg_to_ay_dict(
         project_code_field (str): The ShotGrid project code field.
         custom_attribs_map (dict): Dictionary that maps names of attributes in
             AYON to ShotGrid equivalents.
+        default_task_type (str): The default task type to use if none is found.
     """
     ay_entity_type = "folder"
     task_type = None
@@ -120,9 +121,10 @@ def _sg_to_ay_dict(
         ay_entity_type = "task"
         if not sg_entity["step"]:
             log.warning(
-                f"Task {sg_entity} has no Pipeline Step assigned."
+                f"Task {sg_entity} has no Pipeline Step assigned. "
+                "Task type set from settings."
             )
-            task_type = None
+            task_type = default_task_type
         else:
             task_type = sg_entity["step"]["name"]
 
@@ -133,9 +135,8 @@ def _sg_to_ay_dict(
 
         if label:
             name = slugify_string(label)
-        elif task_type:
+        else:
             name = slugify_string(task_type)
-
 
     elif sg_entity["type"] == "Project":
         name = slugify_string(sg_entity[project_code_field])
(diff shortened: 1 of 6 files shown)


MilaKudr:
During manual sync I got these errors:

  File "C:\Users\Mila\code\ayon-shotgrid\services\processor\processor\processor.py", line 243, in start_processing
    handler.process_event(
  File "C:\Users\Mila\code\ayon-shotgrid\services\processor\processor\handlers\sync_projects.py", line 34, in process_event
    hub.synchronize_projects(source=sync_source)
  File "C:\Users\Mila\code\ayon-shotgrid\services\shotgrid_common\ayon_shotgrid_hub\__init__.py", line 277, in synchronize_projects
    match_shotgrid_hierarchy_in_ayon(
  File "C:\Users\Mila\code\ayon-shotgrid\services\shotgrid_common\ayon_shotgrid_hub\match_shotgrid_hierarchy_in_ayon.py", line 118, in match_shotgrid_hierarchy_in_ayon
    ay_entity = _create_new_entity(
  File "C:\Users\Mila\code\ayon-shotgrid\services\shotgrid_common\ayon_shotgrid_hub\match_shotgrid_hierarchy_in_ayon.py", line 259, in _create_new_entity
    sg_ay_dict["task_type"],
KeyError: 'task_type'
2024-09-16 10:25:53.128 | processor | start_processing | INFO: Event has been processed... setting to finished!

fabiaserra:
is this ignoring them or actually syncing them? the title says it's ignoring but the description says:
> This will change will not raise the missing key and therefore it will allow to sync tasks into project

What's the intended behaviour? if you don't set a task type on the Ayon entity what does it happen?

jakubjezek001:
> is this ignoring them or actually syncing them? the title says it's ignoring but the description says:
> 
> > This will change will not raise the missing key and therefore it will allow to sync tasks into project
> 
> What's the intended behaviour? if you don't set a task type on the Ayon entity what does it happen?

Good point, @fabiaserra. It should ignore them and not produce an error message. We had a case where a client was using SG for development ticket submission. These tasks were in the CG project but didn't have any steps (task types).

  

There are two solutions:

  

1. Ignore them so they aren't synchronized into AYON at all.
2. Assign them a default task type that AYON offers.
3. Offer a settings toggle to choose between the two options mentioned above.

  

@fabiaserra or @robin-ynput, do you have an opinion on this?

fabiaserra:
> There are two solutions:
> 
> 1. Ignore them so they aren't synchronized into AYON at all.
> 2. Assign them a default task type that AYON offers.
> 3. Offer a settings toggle to choose between the two options mentioned above.
> 
> @fabiaserra or @robin-ynput, do you have an opinion on this?

Having the settings control is nice but I think most people are confused with the amount of settings control and it kind of beats the purpose. I think I would probably expect them to sync and use the task name for type in order to sync from SG to AYON and for AYON to SG if there's no step that exists for that type just not set the step when creating the task

robin-ynput:
> > is this ignoring them or actually syncing them? the title says it's ignoring but the description says:
> > > This will change will not raise the missing key and therefore it will allow to sync tasks into project
> > 
> > 
> > What's the intended behaviour? if you don't set a task type on the Ayon entity what does it happen?
> 
> Good point, @fabiaserra. It should ignore them and not produce an error message. We had a case where a client was using SG for development ticket submission. These tasks were in the CG project but didn't have any steps (task types).
> 
> There are two solutions:
> 
> 1. Ignore them so they aren't synchronized into AYON at all.
> 2. Assign them a default task type that AYON offers.
> 3. Offer a settings toggle to choose between the two options mentioned above.
> 
> @fabiaserra or @robin-ynput, do you have an opinion on this?

My understanding is that while `Pipeline Step` are optional in SG, `Task Type` are mandatory in AYON.
So I'd recommend to always sync tasks but either set them as `Generic` or a new `Unknown` type in AYON.

mkolar:
Yeah, if pipeline steps are optional in SG, we need to make a reasonable fallback choice. I like the `unknown` type actually.

jakubjezek001:
> Yeah, if pipeline steps are optional in SG, we need to make a reasonable fallback choice. I like the `unknown` type actually.

Here is the solution added to the current state of this PR. Settings are having new option under Compatibility Settings section for defining a default task type. 
image

MilaKudr (approved):
Works well, tasks are created without any errors on the server.
Screenshot 2024-12-04 095322
Screenshot 2024-12-04 095413

MilaKudr (approved):
Works well, tasks are created without any errors on the server.
Screenshot 2024-12-04 095322
Screenshot 2024-12-04 095413

iLLiCiTiT (approved):

services/shotgrid_common/utils.py
```diff
@@ -119,15 +121,19 @@ def _sg_to_ay_dict(
     if sg_entity["type"] == "Task":
         ay_entity_type = "task"
         if not sg_entity["step"]:
-            raise ValueError(
-                f"Task {sg_entity} has no Pipeline Step assigned."
+            log.warning(
```
iLLiCiTiT, line 124:
```suggestion
            log.debug(
```

PR #134, opened 2024-10-02
Chore: Update versions during create package

iLLiCiTiT:
## Changelog Description
Update version in services `pyproject.toml` during `create_package.py` logic.

## Testing notes:
1. Change version in `package.py`.
2. Run `python create_package.py`.
3. All `pyproject.toml` files should have changed the version to the same as package.py has.


jakubjezek001 (approved):
perfect, thanks

PR #145, opened 2024-10-22
Sync: Prepare for ayon api change

iLLiCiTiT:
## Changelog Description
Use explicit kwargs for all arguments in `EntityHub`.

## Additional info
It is planned to change order of arguments in EntityHub ayon-python-api, this PR just makes sure nothing breaks when the update happens.

## Testing notes:
1. Nothing should change really.
2. Sync should work as did before.


jakubjezek001 (approved):
lgtm

PR #147, opened 2024-11-14
Remodeling settings for Re-parenting of SG entity types for AYON hierarchy

jakubjezek001:
This is just a mockup the settings for presentation of an idea for new settings, which could be used in #129 

Here are some use cases:
### 1. Relocate root folder for SG Asset Entity types

```
├── build 
│   └── assets 
│       └── Prop 
│           ├── AssetProp01
│           └── AssetProp02
│       └── Character
│           ├── AsseCharacter01
│           └── AssetCharacter02
│       └── Environment
│           ├── AssetEnvironment01
│           └── AssetEnvironment2
```
image

### 2. Type grouping for Sequences and Shots
```
├── sequences
|    ├── SQ_010
|    └── SQ_020
├── shots
|    ├── SH_0010
|    └── SH_0020
```
image

## Main Features of the Settings:
1. Each feature has its own category that can be activated. This helps in better validating conflicting settings configurations. For example, "Type Grouping" should not be activated if "Root Relocation Presets" is activated, and vice versa.
2. The parent name input supports Anatomy formatting tokens. Proper documentation is needed, so users should have access to a cheat sheet.

### DIFF
--- a/server/settings/main.py
+++ b/server/settings/main.py
@@ -127,15 +127,6 @@ class AttributesMappingModel(BaseSettingsModel):
         enum_resolver=default_shotgrid_entities
     )
 
-
-class FolderLocationsModel(BaseSettingsModel):
-    """AYON folders to store separate of SG types"""
-    asset_folder: str = SettingsField(title="Assets", default="assets")
-    sequence_folder: str = SettingsField(title="Sequences",
-                                         default="sequences")
-    shot_folder: str = SettingsField(title="Shots", default="shots")
-
-
 class FolderReparentingParentsModel(BaseSettingsModel):
     sg_entity_type: str = SettingsField(
         "Asset",
@@ -231,14 +222,6 @@ class ShotgridCompatibilitySettings(BaseSettingsModel):
         ),
     )
 
-    folder_locations: FolderLocationsModel = SettingsField(
-        title="Folder locations",
-        default_factory=FolderLocationsModel,
-        description=(
-            "Locations of AYON folders matching to SG types."
-            "Eg. where SG assets will be stored, where SG shots and sequences."
-        ),
-    )
     folder_parenting: FolderReparentingModel = SettingsField(
         title="Folder re-parenting",
         default_factory=FolderReparentingModel,
--- a/server/settings/main.py
+++ b/server/settings/main.py
@@ -40,7 +40,6 @@ def default_shotgrid_reparenting_entities():
         "Episode",
         "Sequence",
         "Shot",
-        "AssetCategory",
         "Asset",
     ]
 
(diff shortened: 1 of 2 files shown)


jakubjezek001:
@dee-ynput @kalisp  here is the improved model with better examples in the description.

tweak-wtf:
i don't quite get the difference between `Root relocation` and `Type grouping`. which issue does this try to solve?

in my monkey brain we'd only need one of these as the whole point of this is to follow SG hierarchy tracking under a custom AYON folder.
having said that, in the screen showing asset type sync we would want to remove `Parents-3` from the config when using an SG asset hierarchy tracking `AssetType->Asset`. i'd understand this to end up doubling the AssetCategory assuming the SG tracking `AssetType->Asset` would be applied in AYON hierarchy `/build/assets/{sg_asset[type]}` resulting in `/build/assets/{sg_asset[type]}/AssetType/Asset` where `{sg_asset[type]}` and `AssetType` would hold the same string.
in general when talking about use cases we should also note to what kind of SG Hierarchy Tracking settings we're referring to.

additionally the example given for `type grouping` to me would break the SG tracking when using `Sequence->Shot`. Because SG is set to track `Sequence->Shot` but in AYON (with the shown config) we would place them in different folders.
SO that's what i meant with validating the "correct" configs might be a bit tricky because imho we shouldn't be able to break SG tracking settings. however if a user configured SG Shots and Sequences to have flat Hierarchy Tracking this would again make perfect sense.

generally i believe we would only want to configure the root tracking items from SG (e.g. Sequence in `Sequence->Shot` or AssetType in `AssetType->Asset`) and entities with flat tracking settings.

jakubjezek001:
> i don't quite get the difference between `Root relocation` and `Type grouping`. which issue does this try to solve?

We evolved the design from the previous proposal, where all presets were in the same category, and each preset had a Reparenting type enumerator. This enumerator offered options like `Root relocate` or `Type grouping`. After brainstorming with @kalisp, we decided to split these presets into separate categories, each activatable individually. This approach makes it easier to ensure that only one is used at a time. Both cannot be active simultaneously, as this would collapse the hierarchy. We wanted to avoid a situation where someone might accidentally create presets with both reparenting types.

> in my monkey brain we'd only need one of these as the whole point of this is to follow SG hierarchy tracking under a custom AYON folder. having said that, in the screen showing asset type sync we would want to remove `Parents-3` from the config when using an SG asset hierarchy tracking `AssetType->Asset`. i'd understand this to end up doubling the AssetCategory assuming the SG tracking `AssetType->Asset` would be applied in AYON hierarchy `/build/assets/{sg_asset[type]}` resulting in `/build/assets/{sg_asset[type]}/AssetType/Asset` where `{sg_asset[type]}` and `AssetType` would hold the same string. in general when talking about use cases we should also note to what kind of SG Hierarchy Tracking settings we're referring to.

Yes, you can skip the `Parents-3` part. The system should be flexible enough for that. We haven't been using the AssetType entity yet, and we don't need to because the SG spreadsheet allows grouping by asset type (AssetCategory). The AssetCategory entity type is currently created by default in AYON, and there's no way to disable it. This will be changed by [this PR](https://github.com/ynput/ayon-shotgrid/pull/129). In your example, you would need to set the Root relocate preset and filter it by `AssetType`. The Asset will automatically be kept under its originally set project tracking hierarchy attribute.

> additionally the example given for `type grouping` to me would break the SG tracking when using `Sequence->Shot`. Because SG is set to track `Sequence->Shot` but in AYON (with the shown config) we would place them in different folders. SO that's what i meant with validating the "correct" configs might be a bit tricky because imho we shouldn't be able to break SG tracking settings. however if a user configured SG Shots and Sequences to have flat Hierarchy Tracking this would again make perfect sense.

This wouldn't make sense for you usecase. This applies to the original brief of other client. They need all entities to be grouped by type and to breake whole project tracking hierarchy attribute definition. The original hierarchy should only be kepth in SG.
 
> generally i believe we would only want to configure the root tracking items from SG (e.g. Sequence in `Sequence->Shot` or AssetType in `AssetType->Asset`) and entities with flat tracking settings.

Yes

PR #157, opened 2025-01-09
Bugfix/156 crud error 10 read failed for entity type task

### LINKED ISSUE #156: CRUD ERROR #10: Read failed for entity type [Task]
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
ported from [discord](https://discord.com/channels/517362899170230292/563751989075378201/1326600251762540584):
> I was having an issue with publishing from Harmony 24 into Ayon/Shotgrid. 
> This was for the CRUD error I was getting
> Image

## Expected Behavior

..

### Steps To Reproduce:

.

### Version
Version of the server, ayon-core and the addon that's affected

- shotgrid 0.4.6 addon

### What platform were you running when you found the bug?
Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

You can also provide any additional information relevant to the issue.

### Relevant log output:

```shell

```

MustafaJafar:
## Changelog Description
resolve #156

This solution was ported from [discord](https://discord.com/channels/517362899170230292/563751989075378201/1326603290691108864).

## Testing notes:
1. publishing from harmony should work fine.

### DIFF
--- a/client/ayon_shotgrid/plugins/publish/collect_shotgrid_entities.py
+++ b/client/ayon_shotgrid/plugins/publish/collect_shotgrid_entities.py
@@ -162,12 +162,17 @@ class CollectShotgridEntities(pyblish.api.ContextPlugin):
         if not filters:
             return sg_tasks_by_asset_id
 
-        sg_tasks = sg_session.find(
-            "Task",
-            filters=filters,
-            fields=["content", "entity"],
-            filter_operator="any"
-        )
+        try:
+            sg_tasks = sg_session.find(
+                "Task",
+                filters=filters,
+                fields=["content", "entity"],
+                filter_operator="any"
+            )
+        except Exception as e:
+            self.log.debug(f"An error occurred: {e}")
+            return sg_tasks_by_asset_id
+
         for sg_task in sg_tasks:
             parent_id = sg_task["entity"]["id"]
             task_name = sg_task["content"]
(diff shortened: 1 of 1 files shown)


BigRoy:
Although I can see how this circumvents an error - it would be good to actually receive the full stack trace and error log of the original error because this "try..except" feels like it's catching a case we do actually want to resolve more explicitly. The question becomes _why_ does the task type filter fail for example? :) most likely this is because the task type is undefined or some value that that filter does not support?

MustafaJafar:
> Although I can see how this circumvents an error - it would be good to actually receive the full stack trace and error log of the original error because this "try..except" feels like it's catching a case we do actually want to resolve more explicitly. The question becomes _why_ does the task type filter fail for example? :) most likely this is because the task type is undefined or some value that that filter does not support?
> 
> I mean, there's a stack trace in the issue - but we'd best get an idea of the actual values passed to the database that failed the operation.

great. let's keep this PR in draft mode for future reference.

kahwatikid:
> Although I can see how this circumvents an error - it would be good to actually receive the full stack trace and error log of the original error because this "try..except" feels like it's catching a case we do actually want to resolve more explicitly. The question becomes _why_ does the task type filter fail for example? :) most likely this is because the task type is undefined or some value that that filter does not support?
> 
> I mean, there's a stack trace in the issue - but we'd best get an idea of the actual values passed to the database that failed the operation.

I would love to get that log for you, but I wasn't able to find that data.  After extensively trying to identify what is "unknown", this workaround was good enough for now.  If you have a method where i can get you that information, I'd be happy to help.

MustafaJafar:
Closing this PR for being inactive for too long.

PR #163, opened 2025-01-22
Services: Remove unnecessary dependencies

iLLiCiTiT:
## Changelog Description
Removed `appdirs` and `pydantic` from services dependencies and changed dependencies in service tools from `requirements.txt` to `pyproject.toml`.

I'm not sure what `appdirs` and `pydantic` are defined in dependencies. Pydantic doesn't seems to be used at all, and appdirs are not used by the code itself neither, it is subdependency of shotgun api, but that should be handled automatically.

### Service tools
1. Remove `.venv` in service tools and run `./manage.ps1 createenv`, or `make createenv`.
2. Start all services with service tools.
3. All of them should work as expected.

### Service docker images
1. Build all service docker images.
2. Run them via ash.
3. They should all work as before (transmitter might be crashing because of missing `arrow` aka https://github.com/ynput/ayon-shotgrid/pull/162)


service_tools/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+[project]
+version = "1.0.0"
+name = "shotgrid-service-tools"
+requires-python = ">=3.7"
```
kalisp, line 4: This will not work as `services` require Python `>=3.10,<4.0`

iLLiCiTiT: I don't think it's a big deal, but ok, changed.

kalisp (approved):

PR #167, opened 2025-01-29
Sync reviewable via service

kalisp:
## Changelog Description
This PR removes publish plugins that create SG version and upload reviewable with service implementation triggered by events.
This should make publishing process a bit faster as uploads are not happening from client machines AND could provide us better control of state of reviewable (could be uploaded after some transcoding etc.)

Current state:
publish plugins were removed and replaced by service reacting on events
that way publish should be a bit faster
- creation/update of Version FROM SG to AYON is blocked
SG versions dont contain anything that could be used as product name (reviewArt) or integer version (v001) unless some naming convention would be used
- what should be propagated back from SG
SG Version attribute change produces about 8 events (as viewed_by_current_user) which are irrelevant for us.
some interesting attribute changes could be implemented if requirements will be provided

## Testing notes:
1. update SG service
2. try full project synchronization both ways
3. try publish something that creates reviewable, check that it was created in SG


kalisp:
Current blocker is implementing synchronization of missing/added versions from SG to AYON.

SG Version is missing required properties:
- version number (could be somehow parsed if AND ONLY IF uploaded reviewable has that value)
- product  (there is nothing like `product` in SG)

kalisp:
Putting this as review ready as many people ignore Drafts.

Current state:
- publish plugins were removed and replaced by service reacting on events
    - that way publish should be a bit faster
- creation/update of Version FROM SG to AYON is blocked
    - SG versions dont contain anything that could be used as `product name` (`reviewArt`) or integer version (`v001`) unless some naming convention would be used
    - SG Version attribute change produces about 8 events (as `viewed_by_current_user`) which are irrelevant for us.
    - some interesting attribute changes could be implemented if requirements will be provided

philippe-ynput:
@jakubjezek001 @robin-ynput Can you please look at this PR and see with @kalisp what is needed ? Thanks

robin-ynput:
So I finally got a chance to test this one. Apologies for the delay.

Good news:
    * With these changes I could publish a Render from Nuke with reviews, nothing was asked/expected for me regarding SG
    * New AYON version as properly picked up during `AYON -> SG` sync
    * A new SG version was created with the media attached to it: https://ayon.shotgrid.autodesk.com/page/7019

Not so good news:
   * The code somehow encountered an `IndexError` while `upload_ay_reviewable_to_sg` which gave me a infinite loop trying the reprocess over the same event. Some logs attached (I stopped the daemon manually after 5 attempts):
[version_update_logs.txt](https://github.com/user-attachments/files/19347983/version_update_logs.txt)


Also side thinking:
  * Is it OK to expect that the version's user is sync with SG to process the version sync ? Cannot we go ahread without and simply let the SG version artist unassigned ?
  * From memory it looks like Autodesk hard prevent upload > 5Go, is that something we should check beforehand ?

kalisp:
Couldnt reproduce infinite loop, it should fail after 3 retries 
image
(it is actually difficult to find failing event on `shotgrid.proc` matching to event from `shotgrid.event`, but if you could look for that event, what it is its state?)

I think artist is mandatory as sync of version with user that is not in SG fails.

We should skip uploading such a big (>5GB) file?

robin-ynput (approved):
LGTM with the fixes I mentioned in the comments.
I tried the following:
* Sync existing version from AYON to SG via `project_sync` event
* Sync new version with reviewable via transmitter
* Sync new version without reviewable via transmitter

jakubjezek001 (approved):
versions are synched as expected. It would be good to also add fallback script user for cases when full sync is triggered and a version user does not exists on SG.

PR #177, opened 2025-03-18
Avoid to repopulate all ids and type on AYON->SG sync

robin-ynput:
## Changelog Description

Ensure that triggering `SG -> AYON` manually does not re-populate all of the `shotgridId` and `shotgridType` for AYON entities.

## Testing notes:

1. Ensure this new code is running as a service
2. Have an existing AYON project already in syncing with SG
3. In AYON, create a new Task that would sync with SG
4. Click `AYON -> SG` button in the settings panel
5. Ensure newly added Task get properly set with a `shotgridId` attribute

### DIFF
--- a/services/shotgrid_common/ayon_shotgrid_hub/match_ayon_hierarchy_in_shotgrid.py
+++ b/services/shotgrid_common/ayon_shotgrid_hub/match_ayon_hierarchy_in_shotgrid.py
@@ -225,20 +225,19 @@ def match_ayon_hierarchy_in_shotgrid(
             sg_ay_dicts[sg_entity_id] = sg_ay_dict
             sg_ay_dicts_parents[sg_parent_entity["id"]].add(sg_entity_id)
 
+            # add new Shotgrid ID and type to existing AYON entity
+            ay_entity.attribs.set(
+                SHOTGRID_ID_ATTRIB,
+                sg_entity_id
+            )
+            ay_entity.attribs.set(
+                SHOTGRID_TYPE_ATTRIB,
+                sg_ay_dict["attribs"][SHOTGRID_TYPE_ATTRIB]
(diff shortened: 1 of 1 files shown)


services/shotgrid_common/ayon_shotgrid_hub/match_ayon_hierarchy_in_shotgrid.py
```diff
@@ -225,21 +225,20 @@ def match_ayon_hierarchy_in_shotgrid(
             sg_ay_dicts[sg_entity_id] = sg_ay_dict
             sg_ay_dicts_parents[sg_parent_entity["id"]].add(sg_entity_id)
 
+            # add new Shotgrid ID and type to existing AYON entity
```
BigRoy, line 228: Is this comment now incorrect - because it seems we're not syncing to 'existing' entities in this case at all?

robin-ynput: Hum I think it's fine, we update an existing AYON entity attribute with the ID of the newly created Shotgrid one.

BigRoy: I see, thanks. So it's just the call to `_create_new_entity` that confused me. That sounded like we're generating a "new entity" even though we're not doing so here if I now understand correctly.

jakubjezek001 (approved):
it seems ok.

MustafaJafar (approved):
It works on my side. (I assume the type was set to be `None Set` due to something in my configuration. I'm not sure about it tbh.)
image

MilaKudr (approved):

PR #184, opened 2025-03-25
Fix hierarchy sync for Folder between SG and AYON.

### LINKED ISSUE #175: AY-7593_Broken hierarchy synchronization
### Please describe the issue you have encountered?

Default SG hierarchy structure is not correctly synched into AY
Episode > Sequence > Shot is resulting in broken flat project structure
Root-relocation set to add Folder for Episodes are not creating any folder and all entities are synched with broken flat structure
Root-relocation set to add Folder for Sequences all works in case Episode does not exist on SG project and settings are excluding Episodes from allowed entities

related issue: https://github.com/ynput/ayon-shotgrid/issues/159

### How to replicate the issue?

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1348575232470024252)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-7593](https://app.clickup.com/t/86c2h4ujf)_

robin-ynput:
## Changelog Description

resolve #175 

**Please focus on the hierarchy sync only, current code likely contains a lot of other bugs which will be addressed in different PRs.**

## Testing notes:

Multiple settings have to be done.
1. Leave all of the reparenting settings blank and disabled 
2. Type grouping to group all entities of a common type under a different root `ayon+settings://shotgrid/compatibility_settings/folder_parenting/type_grouping`
3. Root relocate feature to move a hierarchy under a different root `ayon+settings://shotgrid/compatibility_settings/folder_parenting/root_relocate`

For each case, test:
1. sync a new project to a blank 
2. create Episode, Asset, Sequence and Shot from SG and ensure it syncs with AYON through the leecher


robin-ynput:
Question for @jakubjezek001 @philippe-ynput and @kalisp regarding `Asset` and `AssetCategory`. We currently have an inconsistent behaviour:
* When syncing the whole project `(SG -> AYON button)`, the service will attempt to spot an asset category within SG and create a parent `AssetCategory` folder part of the hierarchy
* When syncing via the leecher, none of that happen, we just sync the `Asset` with no `AssetCategory` parent.

Which behaviour should be enforced ?

kalisp:
'Pure' (without any reparenting) full sync SG > AYON throws errors on update of SG entity of `Character` type.
(eg. on smth like 
image
)

image

kalisp:
And I think folder reparenting or grouping is broken for Assets.
Here are descriptions of two usecases https://github.com/ynput/ayon-shotgrid/pull/129#issue-2516248603

Here is setting 
[SG_reparenting_settings.json](https://github.com/user-attachments/files/19471372/SG_reparenting_settings.json)

I wouldn't expect this for `Root relocation`:
image

(problem are paths `assets/char01/char01`, I would expect `assets/Character/char01`)

robin-ynput:
Thanks a lot @kalisp for the feedback. 

The issue with `AssetCategory` error report when syncing to AYON should be fixed.
I'll keep updated for the asset hierarchy one.

jakubjezek001:
> * When syncing the whole project `(SG -> AYON button)`, the service will attempt to spot an asset category within SG and create a parent `AssetCategory` folder part of the hierarchy
> * When syncing via the leecher, none of that happen, we just sync the `Asset` with no `AssetCategory` parent.
> 
> Which behaviour should be enforced ?

There are two ways to solve the issue. One, to enforce the AssetCategory parent way. Another to have for the toggle in settings so anybody can decide which way is better for them. I would assume that enforcing the first way would work for most studios so we could go that way, but not sure if enforcing anything based on our opinion is best approach either. 


> _2- Episode->Sequence parenting_
> 
> Our code reads the SG tracking settings parameters to try to define the hierarchy. The state below is expected to parent the `Sequence` under `Episode` (otherwise they will be parented under `Project`)
> This is a bit blurry on the SG side, because this settings might not be set by default. Also it only enforces the parenting via UI/UX. Nothing prevent a user from defining a parent episode to a Sequence without this stuff enabled. Is it OK to only create a `Episode -> Sequence` hierarchy when this is enabled ? (Not sure I'm clear here).
> 
> The `Sequence -> Shot` parenting seems to come by default, so our client is less likely to miss this one.

We were having the discussion about this and easiest and most anticipated way from community and clients would be to follow the SG project tracking as source of true for entities which are having the parents defined by SG default. In case it is not set by a user - since they can override it in UX, then it should be synchronized into AY project as flat, in case no AY reparenting feature is activated.

kalisp (approved):
I had to remove line  `{sg_type}` from 
image
to limit duplication of `AssetCategory`, but that might not be necessarily bad, unless someone else has strong opinion about it.

philippe-ynput:
@robin-ynput @jakubjezek001 @kalisp Are you ok to merge this so we may prepare a bug-fix release this week ?

jakubjezek001 (changes requested):
Currently I have tried to create new project with all rerooting features OFF. 

Project was created in AY and synced to SG as empty, then I went to Project tracking and set following structure:
`Episode>Sequence>Shot`

Then I was creating it in following order. 
`ep02/seq02/sh020`
and lastly added task to the shot roto(Roto)

It did not happened and I am enclosing the log for more details. 

The services were runned from code. And all was added to Development bundle from the branch code as usual.
[SG_AY_sync.txt](https://github.com/user-attachments/files/19631538/SG_AY_sync.txt)

robin-ynput:
@jakubjezek001 let me know if this works for you.

I can confirm I got the following working on my side:
* Parented `Asset` with `AssetType` grouping enabled synced from `SG` to `AYON` via `project-sync` event
* Parented `Asset` with `AssetType` grouping enabled synced from `SG` to `AYON` via `leecher` event
* Flat `Asset`  (`AssetType` grouping disabled) synced from `SG` to `AYON` via `project-sync` event
* Flat `Asset`  (`AssetType` grouping disabled) synced from `SG` to `AYON` via `leecher` event
* Parented hierarchy for new `Shot/Sequence/Episode` from SG project tracking settings synced from `SG` to `AYON` via `project-sync` event
* Parented hierarchy for new `Shot/Sequence/Episode` from SG project tracking settings synced from `SG` to `AYON` via `leecher` event
* Custom `Root relocation` feature for new entities for  `project-sync` and `leecher` events
* Custom `Type grouping` feature for new entities for  `project-sync` and `leecher` events
* Error when both `Root relocation` and `Type grouping` features are enabled

jakubjezek001 (changes requested):
It seems there is still some problem with AssetCategory parent folder name. In my case the AY folder created for holding SG assets should be named `character` but it is actually having the name of the character typed asset. 

### Current implementation
SG asset: donchuan (type: Character)
AY asset: /assets/donchuan/donchuan

### Expected behavior
AY asset: /assets/character/donchuan

Here are my settings for the project:
[sg_root_relocation_settings.json](https://github.com/user-attachments/files/19645980/sg_root_relocation_settings.json)

Here it seems to be the part of code which might be causing it. https://github.com/ynput/ayon-shotgrid/blob/bugfix/preserve_hierarchy/services/shotgrid_common/utils.py#L536-L553

image

jakubjezek001:
Another issue I have found is retiring of the assets once they are removed from AYON. I was actually removing originally created Episode which was holding a Sequence and shot. But the event only removed the Episode and Sequence with shot is kept in SG. It seems to me that the retiring of entities should happen on all hierarchically related children, right? 

This might be perhaps a bug unrelated to this PR but just want to make sure we know about it.

kalisp:
> It seems there is still some problem with AssetCategory parent folder name. In my case the AY folder created for holding SG assets should be named `character` but it is actually having the name of the character typed asset.

I think it is because of `{shotgrid_type}` PLUS this in settings:
image

robin-ynput:
@kalisp @jakubjezek001 can I please ask you for another round of reviews ?

Chatting with @jakubjezek001 I realized the `AssetCategory` issue was coming from multiple causes:
* `Asset Grouping feature` settings was not exposed at project level
* `Type -> Asset` hierarchy had to be set in SG tracking settings for the whole system to work
* `{shotgrid_type}` placeholder was defaulting to a wrong field when the rule above was not set, also there was a a wrong tooltip suggesting all field could be reached via this placeholder

The situation underlined it was all to complex for our clients, so I changed the whole logic to make it as simple as possible.
1. `Episode -> Sequence` and `Asset Type -> Asset` parenting are handle by SG project tracking settings. That's the only single source of truth.
2. Additional `root reparenting` and `type grouping` options are available as AYON settings but with no placeholder. Values there have to be explicit.

Hopefully this will make the whole easier to configure.

kalisp (approved):
Retested:
- 'pure' sync from SG > AYON
- type parenting
- root relocation

Adding tasks/shots from both ends. 
Syncing from AYON to SG.

jakubjezek001 (approved):
works perfect!

kobflo:
Test this, is working with the following. 

Multiple eps, multiple sequences, shots and tasks.

Important to note that ayon id, and sync status in sg confuses the db and wont update, so needed to blow a bit of the older stuff away. if duplicating anything in SG. Which makes sense i suppose.

Root Reparenting was a bit problematic, but i got type group functional, Yay! 

image
image


Adding shots to an existing sequence after the first population - works!
Adding a new sequence to an existing Episode after the first population - works
Adding a Episode, with seq and shot/tasks after the first population - works


image


Nice work! Thankyou!

robin-ynput:
> Root Reparenting was a bit problematic, but i got type group functional, Yay!

Hey @kobflo, root parenting did not work on your end ? That should be OK..  if you found something not working on your end do you mind to create an issue with that. We'd need your AYON settings, your Flow project tracking settings and as much service logs as possible so we can troubleshoot what's going on.

Issue #188, opened 2025-04-01
Comments under Version entity are not synchronized


jakubjezek001:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
Shotgrid Version entity is having attribute name `description` on our side the attribute is called `description` too but as far as I could find only available attribute at AYON under version is called `comment`. So any SG added descriptions are not synchronizing to our Version comment attribute (labeled Description). 

## Expected Behavior
Description from SG Version entity should be synchronized to AYON version **comment** attribute. 

### Steps To Reproduce:

1. check `ayon+settings://shotgrid/compatibility_settings/custom_attribs_map` and see there is description attribute. 
2. rename the attribute to comment and limit the scope to Version only. 
3. go to SG and add to any description under any version some comment
4. after the attribute is synced via services events synchronization 
5. go to the AYON version under attribute view and see that it was not synchronized

### Version
Version of the server, ayon-core and the addon that's affected

### What platform were you running when you found the bug?
Windows
Linux - Rocky9
Linux - Ubuntu
MacOS

You can also provide any additional information relevant to the issue.

### Relevant log output:

```shell

```

PR #192, opened 2025-04-08
Fix settings field description for comment syncing

kalisp:
Sync loop in transmitter is from AYON to SG (there is no `comment.updated` event topic on AYON side.)


jakubjezek001:
the name of this PR suggesting to me more of the implementation for #188, but this is just about fixing settings description on a field, right @kalisp. This title is quite misleading

jakubjezek001 (approved):
looking good

Issue #181, opened 2025-03-25
AY-7631_Comments are synchronized under wrong entity context


ynbot:
### Please describe the issue you have encountered?

So once Shotgrid user add comment under a version, this is then synchronized under AYON Shot entity. 
On another hand if a comment is added under AYON version the comment is not synchronized at all.

### Additional context:
link to discussion on [Discord](None)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-7631](https://app.clickup.com/t/86c2qk191)_

jakubjezek001:
perhaps duplicity of #172

jakubjezek001:
Just to be more clear. We are talking about activity stream related commentary.

tweak-wtf:
for quickly testing this u can just iterate over the reversed list of `sg_note["note_links"]` in here:
https://github.com/ynput/ayon-shotgrid/blob/929680345f0f7b1684c4c55edce71a87de98389c/services/shotgrid_common/utils.py#L1650

this forces to return the SG version as the parent entity. it works fine for note creation but updating the note will fail with:
```
2025-04-15 12:42:43.797 ERROR: Unable to process handler shotgrid_event
Traceback (most recent call last):
File "/service/processor/processor.py", line 251, in start_processing
handler.process_event(
File "/service/processor/handlers/shotgrid_event.py", line 37, in process_event
hub.react_to_shotgrid_event(sg_payload["meta"])
File "/service/ayon_shotgrid_hub/__init__.py", line 365, in react_to_shotgrid_event
update_ayon_entity_from_sg_event(
File "/service/ayon_shotgrid_hub/update_from_shotgrid.py", line 376, in update_ayon_entity_from_sg_event
handle_comment(sg_ay_dict, sg_session, ayon_entity_hub)
File "/service/utils.py", line 1592, in handle_comment
ay_activity_id = _update_comment(
^^^^^^^^^^^^^^^^
File "/service/utils.py", line 1626, in _update_comment
updated_origin["name"] = ay_parent_entity["name"]
~~~~~~~~~~~~~~~~^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/ayon_api/entity_hub.py", line 1527, in __getitem__
return getattr(self, item)
^^^^^^^^^^^^^^^^^^^
File "/usr/local/lib/python3.11/site-packages/ayon_api/entity_hub.py", line 1964, in get_name
raise NotImplementedError(
[...]

PR #197, opened 2025-04-15
AY-7631 Fix Flow notes from Version not parented properly to AYON.

rdelillo:
## Changelog Description

resolve: #181 

It looks like there is a bug in SG where Note made from the ApiUser do not show up in the Activity page.
E.g. when a user create a comment in AYON to a specific version, it gets reported in Flow but does not appear in the Activity page, of the version.. however I can see it properly in the overhall Note page.
Adding the shot entity in the `note_links` the same way Flow is doing it does not solve anything.. this looks like a bug to me.

image
image

## Testing notes:
1. From an auto-sync project between Flow and AYON
2a. Create a note for a Version in the Activity page and ensure it gets synced to the equivalent Version in AYON
2b. Create a comment in AYON from an unsynced user and ensure it gets created in Flow (unassigned)

### DIFF
--- a/services/shotgrid_common/ayon_shotgrid_hub/__init__.py
+++ b/services/shotgrid_common/ayon_shotgrid_hub/__init__.py
@@ -497,18 +497,18 @@ class AyonShotgridHub:
                     sg_user_id_by_user_name, ayon_username)
 
                 if sg_user_id < 0:
-                    self.log.warning(
-                        f"Author {ayon_username} is not "
-                        "synchronized to SG, skipping comment"
+                    self.log.debug(
+                        f"Author {ayon_username} is not synchronized "
+                        "to SG, comment will be left unassigned."
                     )
-                    continue
+                    sg_user_id = None
 
                 self._create_sg_note(
                     self.project_name,
                     entity_dict,
                     activity,
-                    sg_user_id,
-                    sg_user_id_by_user_name
+                    sg_user_id_by_user_name,
+                    author_sg_id=sg_user_id,
                 )
             else:
                 sg_update_data = {}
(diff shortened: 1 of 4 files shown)


tweak-wtf:
concerning Shot/Asset, Task, Version Comment Sync:
- AYON -> SG
  - i can second everything u wrote on the PR description.
  - asked Autodesk Support about this. lets see.
  - additionally this also happens for synced users
- SG -> AYON
  - Shot/Asset Comments: syncs as expected
  - Version Comments: syncs as expected
  - Task Comments: this is odd on SG's side
    - even if writing a note with selected task SG will only link the asset/shot (parent entity) and not its task. another SG bug?

concerning SG Note replys:
- these currently don't get synced to AYON
- SG has a specific `Reply` entity type that's different from `Note`
  - comes with its own SG event type `Shotgun_Reply_New` which we'd need to handle to support this
- afaik AYON also doesn't allow for comment replys
  - could we use markdown quotes (`>`) for this similiar to how it's done on GitHub?
    - AYON would format it nicely 🤷

kalisp (approved):
Tested notes on Version, from both direction.
Can confirm issue with Activity feed in SG, googling pointed in issue of permissions, but it didnt look it that way in SG/couldnt make it work.

tweak-wtf:
> Tested notes on Version, from both direction. Can confirm issue with Activity feed in SG, googling pointed in issue of permissions, but it didnt look it that way in SG/couldnt make it work.

yeah Autodesk just confirmed to me that it is indeed a bug on their end

tweak-wtf:
i made a thing for syncing SG note comments to AYON in https://github.com/ynput/ayon-shotgrid/pull/198

PR #202, opened 2025-04-22
Bi-Directional attachments sync on notes/comments

tweak-wtf:
## Changelog Description
Supports bi-directional note attachments sync between AYON and SG.

- gets uploaded attachments and temporarily downloads them locally
- downloaded attachment is uploaded to the other CMS
- removing an attachment is reflected in the opponent CMS

## Testing Notes
- it's not possible to add additional attachments after a note is created neither in AYON nor in SG

### TODOs
- [x] sync SG attachments to AYON
- [x] sync AYON attachments back to SG
- [x] handle bi-directional attachment deletion
- [ ] listen to SG event also when only the `attachments` field is updated

### DIFF
--- a/services/shotgrid_common/utils.py
+++ b/services/shotgrid_common/utils.py
@@ -1782,7 +1782,6 @@ def _add_comment(
             attachment_paths = [atch["local_path"] for atch in sg_note["attachments"]]
             for path in attachment_paths:
                 mime_type, _ = mimetypes.guess_type(path)
-                log.info(f"{mime_type = }")
                 headers = {
                     "Content-Type": mime_type,
                     "x-file-name": os.path.basename(path),
--- a/services/shotgrid_common/utils.py
+++ b/services/shotgrid_common/utils.py
@@ -1668,11 +1668,19 @@ def _get_sg_note(sg_note_id, sg_session):
     if sg_note.get("attachments"):
         # download attachments locally
         for attachment in sg_note["attachments"]:
-            with tempfile.NamedTemporaryFile(
-                delete=False,
-                suffix=attachment["name"]
-            ) as tmp_file:
-                attachment["local_path"] = sg_session.download_attachment(attachment, file_path=tmp_file.name)
+            try:
+                with tempfile.NamedTemporaryFile(
+                    delete=False,
+                    suffix=attachment["name"]
+                ) as tmp_file:
+                    attachment["local_path"] = sg_session.download_attachment(attachment, file_path=tmp_file.name)
+            except Exception as error:
+                log.warning(
+                    "Could not download note attachment for %r (attachment: %r): %r.",
+                    sg_note,
+                    attachment,
+                    error
+                )
     return sg_note, sg_note_id
 
 
--- a/services/shotgrid_common/utils.py
+++ b/services/shotgrid_common/utils.py
@@ -1575,6 +1575,7 @@ def handle_comment(sg_ay_dict, sg_session, entity_hub):
 
     if not ayon_comment:
         ay_activity_id = _add_comment(
+            sg_session,
             project_name,
             ay_parent_entity["id"],
             ay_parent_entity["entity_type"],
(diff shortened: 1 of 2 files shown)


tweak-wtf:
note to self: deleting/updating attachments SG-AY works now but only if the SG note body also is changed. only deleting the attachment doesn't cause our services to trigger.
probably need to listen to an additional SG event type here?

on that note: if i'm not mistaking it's impossible to update/add note attachments in SG after it was created. a user can only remove the attachment but not add aditional ones

services/shotgrid_common/utils.py
```diff
@@ -1638,12 +1639,36 @@ def _update_comment(
...
+                #     endpoint=f"projects/{project_name}/files/{ay_atchmt['id']}"
+                # )
+                pass
+
+        for sg_atchmt in sg_atchmts:
+            # we can assume only new attachments here bc we popped the already existing ones
+            if atch_id := _handle_attachment(sg_session, sg_atchmt, project_name):
+                file_ids.append(atch_id)
+
+    ayon_api.update_activity(   #! gotta check if this causes notes to be updated everytime
```
tweak-wtf, line 1665: on that note i noticed that after a new note entity is created in SG the update logic also runs. is that expected?

MilaKudr:
Sync of attachments works without issue from SG to Ayon. Sync from Ayon to SG only works partially, as only text notes are uploaded, the picture attachment is deleted after uploading.
Screenshot 2025-04-23 152709
Screenshot 2025-04-23 152653
Screenshot 2025-04-23 152733

MilaKudr:
Adding annotations seems to work, too. 
Screenshot 2025-04-23 161253

tweak-wtf:
> Sync of attachments works without issue from SG to Ayon. Sync from Ayon to SG only works partially, as only text notes are uploaded, the picture attachment is deleted after uploading.

hm that's a bit weird tho. can't reproduce this on my end :/
i get attachment added in AYON synced over to SG also for versions. also they show up at the version. details page with preview

maybe... i'm using the general file attachment (the little paperclip icon) since i don't have the power pack available on my ayon instance. could this be a reason? did u use the annotations from power pack? would that even make a difference?

MilaKudr:
> maybe... i'm using the general file attachment (the little paperclip icon) since i don't have the power pack available on my ayon instance. could this be a reason? did u use the annotations from power pack? would that even make a difference?
> 

Aaaah, yes, I was using the annotations from the Power Pack. (there was probably a misunderstanding on my side) I will retest it with the attachments later today. The question is what this PR should include. Should it be for attachments only, or should  annotations be included as well?

tweak-wtf:
> I will retest it with the attachments later today.

thanks mila :) i also added 2 debug loggings to help track this. 

> The question is what this PR should include. Should it be for attachments only, or should  annotations be included as well?

yes, it should definitely also support the powerpack annotations. i just currently have no way of testing this since i cant install it on my dev ayon instance. so to help myself i used the "normal" file attachment assuming itdl would be somewhat similiar

MilaKudr:
> deleting/updating attachments SG-AY works now but only if the SG note body also is changed. only deleting the attachment doesn't cause our services to trigger. probably need to listen to an additional SG event type here?
> 
> on that note: if i'm not mistaking it's impossible to update/add note attachments in SG after it was created. a user can only remove the attachment but not add aditional ones
> 
> EDIT: it's not possible to add additional attachments after a note is created neither in AYON nor in SG

just confirm that all of these is true

services/shotgrid_common/utils.py
rdelillo: Yes I believe so, SG creates the new row for the note entity in its table, then set the values. So we react to both events.

rdelillo (commented):
Confirmed it works for me, both ways. Thanks a lot for that !
* [X] Create and report the attachement when creating a note from Flow to AYON
* [X] Create and report the attachement when creating a note from AYON to Flow
* [X] Delete the attachement from Flow to AYON *if the note body is edited*
* [X] Delete the attachement from AYON to Flow
* [X] Confirmed attachements seems uneditable once note/comment in created in Flow/AYON

> deleting/updating attachments SG-AY works now but only if the SG note body also is changed. only deleting the attachment doesn't cause our services to trigger. probably need to listen to an additional SG event type here?

You're right, from what I can see this would be a `Shotgun_Attachment_Retirement` event. We would need this to somehow behave like a `Shotgun_Note_Update`.

rdelillo (approved):

PR #204, opened 2025-04-22
AY-7598 Fix sync task statuses

### LINKED ISSUE #176: AY-7598_Broken Statuses synchronization AY to SG
### Please describe the issue you have encountered?

Problem:
There is an issue with the autosync from AY to SG. When a task status is changed in SG, it syncs to AY. However, changes made in AY do not sync back to SG.

Details:
A shot with tasks and a user was synced from SG to AY.
Status changes in SG successfully sync to AY.
Status changes in AY do not trigger any event or sync back to SG.

Conclusion:
The mapping and communication from SG to AY are correct, as evidenced by successful syncing and event logging. The issue lies in the lack of event triggering when changes are made in AY.

### How to replicate the issue?

_No response_

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1348575232470024252/1349044720411738224)
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-7598](https://app.clickup.com/t/86c2hxtbc)_

rdelillo:
## Changelog Description

resolve #176 
Changes:
* Rename statuses when syncing anatomy from SG if the status name already exists (confusion "In Progress")
* Ensure statuses are synced for all entities/folder type
    *  Manual push sync (AYON -> Flow and Flow->AYON)
    * Automated sync

## Testing notes:
1. Run the services
2. Ensure all of the statuses sync when applicable

### DIFF
--- a/services/shotgrid_common/ayon_shotgrid_hub/match_ayon_hierarchy_in_shotgrid.py
+++ b/services/shotgrid_common/ayon_shotgrid_hub/match_ayon_hierarchy_in_shotgrid.py
@@ -102,6 +102,7 @@ def match_ayon_hierarchy_in_shotgrid(
         status.name: status.short_name
         for status in  entity_hub.project_entity.statuses
     }
+    all_sg_statuses = {}
 
     while ay_entity_deck:
         (sg_ay_parent_entity, ay_entity) = ay_entity_deck.popleft()
@@ -200,7 +201,12 @@ def match_ayon_hierarchy_in_shotgrid(
                     ay_project_sync_status = "Failed"
 
             attrib_values = {}
-            sg_statuses = get_sg_statuses(sg_session, sg_entity_type)
+            if sg_entity_type in all_sg_statuses:
+                sg_statuses = all_sg_statuses[sg_entity_type]
+            else:
+                sg_statuses = get_sg_statuses(sg_session, sg_entity_type)
+                all_sg_statuses[sg_entity_type] = sg_statuses.copy()
+
             short_name = ay_statuses.get(ay_entity.status)
             if short_name in sg_statuses:
                 attrib_values["status"] = short_name
--- a/services/shotgrid_common/ayon_shotgrid_hub/match_ayon_hierarchy_in_shotgrid.py
+++ b/services/shotgrid_common/ayon_shotgrid_hub/match_ayon_hierarchy_in_shotgrid.py
@@ -20,7 +20,8 @@ from utils import (
     get_sg_entity_as_ay_dict,
     get_sg_custom_attributes_data,
     create_new_sg_entity,
-    upload_ay_reviewable_to_sg
+    upload_ay_reviewable_to_sg,
+    get_sg_statuses,
 )
 
 from utils import get_logger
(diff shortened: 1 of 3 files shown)


services/shotgrid_common/ayon_shotgrid_hub/update_from_ayon.py
```diff
@@ -226,10 +226,13 @@ def update_sg_entity_from_ayon_event(
         # Otherwise it's a tag/status update
         elif ayon_event["topic"].endswith("status_changed"):
             sg_statuses = get_sg_statuses(sg_session, sg_entity_type)
-            for sg_status_code, sg_status_name in sg_statuses.items():
-                if new_attribs.lower() == sg_status_name.lower():
```
rdelillo, line 230: relying on the name is not accurate, should rely on the short_name instead.

services/shotgrid_common/ayon_shotgrid_hub/match_ayon_hierarchy_in_shotgrid.py
```diff
@@ -192,10 +193,21 @@ def match_ayon_hierarchy_in_shotgrid(
                     )
                     ay_project_sync_status = "Failed"
 
-            # Update SG entity custom attributes with AYON data
+            attrib_values = {}
+            ay_statuses = {
```
kalisp, line 197: Shouldn't this be outside of the loop?
I think it definitely could for AYON as you are always looking for statuses that are on `project_entity`.

There are different statuses per `sg_entity_type`? Couldnt this be cached somehow? (Not sure if it actually matters, if that query is fast enough, it just feels "weird" ;))

services/shotgrid_common/utils.py
```diff
@@ -403,13 +403,17 @@ def create_sg_entities_in_ay(
     project_entity.folder_types = new_folder_types
 
     # Add ShotGrid Statuses to AYON Project Entity
-    ay_status_codes = [s.short_name.lower() for s in list(project_entity.statuses)]
+    ay_statuses = {s.short_name.lower(): s.name.lower() for s in list(project_entity.statuses)}
```
kalisp, line 406: There are some places in this diff where Ruff will be complaining because line widths (it should be 80, I think)

services/shotgrid_common/ayon_shotgrid_hub/match_ayon_hierarchy_in_shotgrid.py
rdelillo: No you're right, this is a good call.

services/shotgrid_common/utils.py
rdelillo: Yeah, I'm surprised it does not:
https://github.com/ynput/ayon-shotgrid/actions/runs/14708184716/job/41273491262?pr=204

services/shotgrid_common/ayon_shotgrid_hub/match_ayon_hierarchy_in_shotgrid.py
kalisp: Would there be caching added even for `get_sg_statuses`? Currently you would be asking for statuses for each synchronized item from SG unnecessarily. I am assuming only limited set of `sg_entity_type` but possible large amount of synchronized items. 
(Not sure if `sg_session` doesn't cache it somehow already.)

MilaKudr (approved):
Task statuses are synchronised correctly in both directions. 
Screenshot 2025-04-30 111416
Screenshot 2025-04-30 113937

kalisp (approved):
LGTM, thanks for changes about caching.

PR #206, opened 2025-04-29
[WIP DO NOT REVIEW YET] Enhancement/tests

rdelillo:
## Changelog Description
Brief description of the solution this PR is implementing. This should be only a few sentences and it will be publicly visible in the changelog. You can highlight importance of the change and add links to further information or supporting documentation

Detailed information of the changes made to the product or service, providing an in-depth description of the updates and enhancements. This can include technical information, code examples and anything else needed for the review of the PR.

## Testing notes:
1. start with this step
2. follow this step


service_tools/pyproject.toml
```diff
@@ -8,3 +8,10 @@ dependencies = [
     "ayon-python-api>=1.0.0",
     "shotgun-api3==3.5.1",
 ]
+
+[project.optional-dependencies]
+test = [
+    "pytest",
+    "mock",
+    "pytest-ayon @ git+https://github.com/ynput/pytest-ayon.git@enhancement/split_project_fixture",
```
rdelillo, line 16: TODO:
Change this once https://github.com/ynput/pytest-ayon/pull/2 is merged.

services/shotgrid_common/utils.py
rdelillo, line 1: To explain a bit better the changes here:

* when running via automated tests, `compatibility_settings` are not available. Hence change the logic to get them instead of assuming they are there.
* rework the id stored in the `sg_ay_entity_dicts` container, they used to be the shotgrid ID directly but Mockgun does not guarantee them unique. (e.g. first shot and first sequence created will both be `ID=1`, so I have added the shotgrid type part of the key to guarantee unicity).

MilaKudr (approved):
[
Screenshot 2025-06-05 232929
](url)

PR #211, opened 2025-05-06
Report missing attributes sync for Version.

rdelillo:
## Changelog Description

When moving the `Version` creation sync from plugin to the service, we forgot to report some attributes that used to sync:
* `sg_task`
* `sg_version_type` (when available cause Flow restrict the list choices here)
* `sg_first_frame`
* `sg_last_frame`
* `description`

## Testing notes:
1. Run the services and try to sync a new Version from AYON to Flow
2. Ensure the listed attributes are properly reported

### DIFF
--- a/services/shotgrid_common/utils.py
+++ b/services/shotgrid_common/utils.py
@@ -1958,6 +1958,7 @@ def create_new_sg_entity(
         if not ayon_asset:
             raise ValueError(f"Not found '{folder_id}'")
 
+        # sync version author
         ay_username = ay_entity.data["author"]
         sg_user_id = get_sg_user_id(ay_username)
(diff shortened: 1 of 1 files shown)


kalisp (approved):
Comment got pushed through.

I couldnt find `first_frame` in SG, but its being populated, so hopefully it will turn up in SG.

rdelillo:
> Comment got pushed through.
> 
> I couldnt find `first_frame` in SG, but its being populated, so hopefully it will turn up in SG.

Do you have a path to the Version you created so I can check ?
There is a `frame_start (sg_frame_start)` field and a `First Frame (sg_first_frame)` (the one we update) field in our Flow instance.
Which is quite confusing.

PR #223, opened 2025-05-14
Ignore SG events coming from ourselves.

rdelillo:
## Changelog Description

Our `leecher` service code was not ignoring events coming from us, and supposely ignoring anything from `ApiUser`.

I believe this is wrong, we should only ignore events generated by the AYON<->SG API script.
Everything else is OK to go through.

## Testing notes:

image

1. Ensure that the ApiScript that handle the sync in Flow has is generating events
2. Make some sync from AYON to SG.
3. Ensure that the resulting events are ignored by the leecher
4. Turn off event generation for the ApiUser.

### DIFF
--- a/services/leecher/leecher/listener.py
+++ b/services/leecher/leecher/listener.py
@@ -125,6 +125,11 @@ class ShotgridListener:
             self.log.error("Unable to connect to Shotgrid Instance:")
             raise e
 
+        self.sg_current_api_user = self.sg_session.find_one(
+            "ApiUser",
+            [["firstname", "is", self.sg_script_name]]
+        )
+
         signal.signal(signal.SIGINT, self._signal_teardown_handler)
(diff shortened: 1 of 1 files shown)


services/leecher/leecher/listener.py
```diff
@@ -355,19 +360,24 @@ def start_listening(self):
...
 
         Args:
             event (dict): The Shotgrid Event data.
 
         Returns:
-            bool: True if the event was caused by an API user.
+            bool: True if the event was caused by our API user.
         """
-        # TODO: we have to create specific api user filtering
+        # Ignore events that are comming from ourselves.
```
BigRoy, line 371:
```suggestion
        # Ignore events that are coming from ourselves.
```

kalisp (approved):
Looks better:
image

MilaKudr (approved):

PR #227, opened 2025-05-21
Rework datetime conversion and exception catch.

rdelillo:
## Changelog Description
Adressing technical feedback from @kalisp:
* potential uncaught `TypeError`
* `try-except-else` not that readable

## Testing notes:
1. Sync date and non-date attribute between AYON and Flow
(follow detailed guideline in https://github.com/ynput/ayon-shotgrid/pull/200)


kalisp (approved):
This seems more readable to me.

PR #245, opened 2025-07-11
Added frame_count and frame_range fields when Version is created

kalisp:
## Changelog Description
Fills additional fields:
- frame_count
- frame_range
on Shotgrid Version.

These might be used in various delivery tools.

## Testing notes:
1. synchronize/create new Version
2. check mentioned fields on SG
image


rdelillo (approved):
LGTM, Good stuff !

I was concerned at first that `frame_count` and `frame_range` would be client-specific, but it looks like they are default to Flow's Version entity.
image

MilaKudr:
Also work for me. 
Screenshot 2025-07-14 141238

PR #246, opened 2025-07-24
Don't restrict shotgun_api3 version

antirotor:
## Changelog Description
Shotgun API is bundling *certifi* - using old version can lead to issues while connecting to SG/Flow servers.

`shotgun_api3` is vendoring some libraries including `certifi`. Services using this API are connecting to various Flow/SG servers and if old version of certifi is used, the connection will fail. We can use the latest API  because we can assume that SG instances will use the latest public API too.

## Testing notes:
This is hard to test. It should just work.


rdelillo (approved):

MilaKudr:
I also tested the PR, pyproject.toml files are tweaked, it works for my case.

PR #251, opened 2025-08-08
Fix: Adding MoviePath for farm rendering too

kalisp:
## Changelog Description
`MoviePath` was used only on local rendering.

## Testing notes:
1. use configuration from https://github.com/ynput/ayon-shotgrid/pull/248
2. publish on farm  and locally
3. both should behave same


64qam:
It works, the only thing is that Integrate traits for SG displays: 

```DEBUG: Instance `render` does not have traits
DEBUG: Instance `workfile` does not have traits```
[publish-report-250809-22-31.json](https://github.com/user-attachments/files/21701480/publish-report-250809-22-31.json)

PR #256, opened 2025-08-15
Fix missed break to uses preferred extension

kalisp:
## Changelog Description
Use preferred configured extension if found

## Testing notes:
1. configure `ayon+settings://shotgrid/publish/ExtractMoviePath/profiles/0/repre_names` with multiple representation name
2. check SG that it pulled correct path from correct representation


MilaKudr:
image

MilaKudr (approved):

Issue #255, opened 2025-08-14
Comments synchronization ignores enabled entitites.


kalisp:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

## Current Behavior
Each project could have set up different set of `Enabled Entities` in `ayon+settings://shotgrid/compatibility_settings/shotgrid_enabled_entities`. This seems to not have any impact on comments synch.

## Expected Behavior
I would expect if `Note` is not set in `Enabled Entities` even comments from AYON shouldnt be synchronized to SG.

### Version
0.6.8

### Relevant log output:

```shell

```

PR #266, opened 2025-09-29
Sync comment only if enabled in compatibility settings

tweak-wtf:
adresses https://github.com/ynput/ayon-shotgrid/issues/255

## Changelog Description
Added conditional to check if `Comment` is enabled in compatibility settings and only syncs if enabled.

- in my testing SG->AYON comment sync is already disabled when `Comment` isn't active in compatibility settings

### DIFF
--- a/services/transmitter/transmitter/transmitter.py
+++ b/services/transmitter/transmitter/transmitter.py
@@ -177,7 +177,7 @@ class ShotgridTransmitter:
                 # Run comments sync
                 now_time = arrow.utcnow()
                 sec_diff = (now_time - last_comments_sync).total_seconds()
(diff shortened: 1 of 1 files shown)


kalisp:
I don't think here is correct location. Enabled entities might be different per project, theoretically.  I think it should be lower in `_sync_comments` but before `dispatch_event` on line 301 to stop unnecessary pollution of event queue.

tweak-wtf:
got ya.
only question i'd have is:
will service addons be able to support the "Per Project Bundle" feature? IIRC these addons were the exception

kalisp:
Yeah, I am trying to update to project settings in Transmitter, I should have updated PR today for you to test.

Issue #199, opened 2025-04-16
Implement sync support for Note Reply from Flow


rdelillo:
### Is there an existing issue for this?

- [x] I have searched the existing issues and added correct labels.

### Please describe the feature you have in mind and explain what the current shortcomings are?

Flow defines a separate entity type for answers to `Note` called `Reply`:
![Image](https://github.com/user-attachments/assets/196e9a76-2394-4e57-b3cd-6f75e5899d53)

Those are currently not handled by the synchronization.
Also currently AYON does not have the concept of conversation for its `comment` activities.

### Suggested implementation?

Ideally we'd like `Reply` to be treated same as `Note`. To work-around the issue with AYON not having chat possibilies, maybe we could at leat point to the initial text the comment answers to via a `>` quote syntax.

More info for the initial discussion here: https://github.com/ynput/ayon-shotgrid/pull/197#issuecomment-2809105557

tweak-wtf:
a note on this.
in SG you can only reply to a `Note`. meaning you can't reply on a `Reply`. so nested conversations are impossible.
should come in handy here 🤷 

i believe using markdown quote styling`>` would be our best bet on implementing this without having to add reply functionalities in AYON Activity Stream. they get nicely formatted in AYON as well as in SG.
basically mimicking how GitHub also handles replies in conversations

tweak-wtf:
i'm spending some time on this atm.

found out that unfortunately we can't add custom fields to `Reply` in sg. so we can't persist AYON ID on the sg entity, bummer.
i can sync the reply with citation styling... just for updating/removing replies i really have no clue how to achieve this without being able to persist AYON IDs on SG replies :/

PR #267, opened 2025-09-29
Support for syncing SG `Reply` entity

tweak-wtf:
adresses https://github.com/ynput/ayon-shotgrid/issues/199

## Changelog Description
My take on supporting syncing of `Reply` entities SG->AYON. 

`Reply` is a very special entity type in SG that's not as easy to work with. It doesn't save the project it's connected and also doesn't allow for new custom attributes (like `sg_ayon_id`).
For that reason i didn't find a way to support deleting a reply in AYON when it is deleted in SG.

## Testing notes:
1. build/upload new server settings
2. configure `Reply` as supported entity type in compatibility settings
3. build/run containers
4. write/update a reply in SG

### DIFF
--- a/server/settings/main.py
+++ b/server/settings/main.py
@@ -22,7 +22,8 @@ def default_shotgrid_entities():
         "Asset",
         "Task",
         "Version",
-        "Note"
+        "Note",
+        "Reply",
     ]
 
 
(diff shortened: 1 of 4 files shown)


kalisp (approved):

kalisp:
@tweak-wtf could you please merge origin/develop, resolve conflict and push? I tried to do it, but it seems that I dont have permissions on your fork.

PR #268, opened 2025-09-30
Sync comment only if enabled in compatibility settings

kalisp:
## Changelog Description
This PR provides project based control of comments synchronization and more granular support for synched custom attributes (per project).

## Testing notes:
1. enable/disable Notes in your synchronized project `ayon+settings://shotgrid/compatibility_settings/shotgrid_enabled_entities`
2. experiment also with project based `ayon+settings://shotgrid/compatibility_settings/custom_attribs_map`

### DIFF
--- a/services/transmitter/transmitter/transmitter.py
+++ b/services/transmitter/transmitter/transmitter.py
@@ -177,7 +177,7 @@ class ShotgridTransmitter:
                 # Run comments sync
                 now_time = arrow.utcnow()
                 sec_diff = (now_time - last_comments_sync).total_seconds()
-                if sec_diff > COMMENTS_SYNC_INTERVAL and "Note" in self.sg_enabled_entities:
+                if sec_diff > COMMENTS_SYNC_INTERVAL:
                     self._sync_comments()
 
                 # enrolling only events which were not created by any
@@ -297,6 +297,10 @@ class ShotgridTransmitter:
 
         if activities_after_date is None:
             activities_after_date = now - timedelta(days=5)
+
+        if "Note" not in self.sg_enabled_entities:
+            self.log.warning("Skipping comments sync, 'Note' entity is not enabled.")
+            return
 
         response = ayon_api.dispatch_event(
             SHOTGRID_COMMENTS_TOPIC,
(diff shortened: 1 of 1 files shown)


tweak-wtf:
just gave it a spin checking out the note sync as i don't really use custom attributes in SG or AYON.

i tried having `Note` enabled in projectA and disabled in projectB.
seems the transmitter respects the project override but leecher/processor does not. 

without any project overrides this works as intended

kalisp:
Yes, that true as Mila mentioned ` but only for Ayon to SG direction(eg transmitter). For the SG to Ayon direction, only studio settings are applied (eg leecher/publisher)`.

Sadly we dont have buget/request to implement it from SG. I got carried away at the end of last month where there is a bit leeway  with scheduled work.

I cannot promise I'll get to this now, but I would like to implement it eventually. Question is, if we merge this one and wait for next leg, or we remove it here and wait for whole deal.

kalisp:
> seemed like a tiny fix to me

That sounds familiar, thats why ventured into project overrides and then it snowballed :)

jakubjezek001 (approved):
looking good

services/transmitter/transmitter/transmitter.py
```diff
@@ -357,15 +363,44 @@ def _cleanup_in_progress_comment_events(self) -> bool:
...
         """Finds last successful run of comments synching to SG."""
         finished_events = list(ayon_api.get_events(
-            topics={SHOTGRID_COMMENTS_TOPIC},
-            statuses={"finished"},
-            limit=1,
-            order=ayon_api.SortOrder.descending,
+                topics={SHOTGRID_COMMENTS_TOPIC},
+                statuses={"finished"},
+                limit=1,
+                order=ayon_api.SortOrder.descending,
```
iLLiCiTiT:
```suggestion
            topics={SHOTGRID_COMMENTS_TOPIC},
            statuses={"finished"},
            limit=1,
            order=ayon_api.SortOrder.descending,
```

iLLiCiTiT (approved):

PR #275, opened 2025-11-11
Fix: error when sg_entity couldn't be created

kalisp:
## Changelog Description
It is expected that `create_sg_entity_from_ayon_event` will always return source `ay_entity` even if Shotgrid entity couldn't be created.

Solves:
```
  File "/service/transmitter/transmitter.py", line 223, in start_processing
    hub.react_to_ayon_event(source_event)
  File "/service/ayon_shotgrid_hub/init.py", line 438, in react_to_ayon_event
    update_sg_entity_from_ayon_event(
  File "/service/ayon_shotgrid_hub/update_from_ayon.py", line 251, in update_sg_entity_from_ayon_event
    sg_id = ay_entity.attribs.get("shotgridId")
            ^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'attribs'
```
It should return `ay_entity` which is used to query updated/none `shotgridId`.

In this case it is triggered if Task should be synchronized to SG, but its parent has non integral value in `shotgridId` (not yet sure how this could happen). This doesn't solve original problem, but it helps debugging and manual mitigation.

## Testing notes:
1. to replicate mangle `shotgridId` to any string value
image

### DIFF
--- a/services/shotgrid_common/ayon_shotgrid_hub/update_from_ayon.py
+++ b/services/shotgrid_common/ayon_shotgrid_hub/update_from_ayon.py
@@ -77,15 +77,18 @@ def create_sg_entity_from_ayon_event(
 
     sg_entity = None
 
-    if sg_id and not isinstance(sg_id, int):
-        log.warning(
-            f"Skip SG entity processing from AYON '{ay_entity}'. "
-            f"AYON entity defines an invalid non-integer sg_id '{sg_id}'."
-        )
-        return ay_entity
+    if sg_id is not None:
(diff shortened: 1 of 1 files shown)


rdelillo:
> When I change the SG ID of an entity in Ayon to nonsense, a valid warning appears instead of an unhandled error. I still don't know how to achieve the same scenario from Flow to Ayon.

That's the test. The code changes only apply from AYON to SG so you cannot really test the other way around.

PR #299, opened 2026-06-16
Fix attribute inheritance value detection on AYON -> SG sync.

### LINKED ISSUE #101: Custom attributes in SG set explicitly even for AY hierarchical values
Currently, we are not explicitly populating values into synchronized custom attributes during project synchronization. As a result, if some AYON folder attributes are left unset, they will naturally inherit values from their parent folders. 
image

This leads to SG showing no value and appearing empty. To address this issue, we need to ensure that hierarchical values are resolved during project synchronization. 
image

We should explicitly set default values for synchronized attributes and, during the reverse synchronization of attributes from SG to AY, we must verify if values are inherited hierarchically. If they are, we should ignore them to prevent any unintended local entity overrides.

rdelillo:
## Changelog Description

fixes: #101 
When a new entity was created on AYON with attributes inheriting values from parent, those values were not reported in SG properly.

## Testing notes:

1. Set up some attribute configuration in your addon settings (e.g. resolution or frame range)

Manual project sync:
1. Create a new folder in AYON with inherited attributes
2. Trigger AYON -> SG
3. Ensure new entity is created and values are reported as static values in SG

Automated sync:
1. Configure automated sync based on events
2. Create a new folder in AYON with inherited attributes
3. Ensure new entity is created and values are reported as static values in SG

### DIFF
--- a/services/shotgrid_common/utils.py
+++ b/services/shotgrid_common/utils.py
@@ -2228,15 +2228,20 @@ def create_new_sg_entity(
             if ay_status and ay_status in get_sg_statuses(sg_session, sg_type):
                 data["sg_status_list"] = ay_status
 
+    attr_values = ay_entity.attribs.to_dict()
     # Fill up data with any extra attributes from AYON we want to sync to SG
-    # Use folder entity to get inherited attribute values.
-    folder_entity = ayon_api.get_folder_by_id(
-        ay_entity.project_name,
-        ay_entity.id,
-    )
+    # If some values are None, that could mean the attribute is inherited from the parent.
+    # Use folder entity to get inherited attribute values then.
+    if None in attr_values:
+        folder_entity = ayon_api.get_folder_by_id(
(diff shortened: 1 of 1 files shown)


services/shotgrid_common/utils.py
```diff
@@ -2228,10 +2228,20 @@ def create_new_sg_entity(
             if ay_status and ay_status in get_sg_statuses(sg_session, sg_type):
                 data["sg_status_list"] = ay_status
 
+    attr_values = ay_entity.attribs.to_dict()
     # Fill up data with any extra attributes from AYON we want to sync to SG
+    # If some values are None, that could mean the attribute is inherited from the parent.
+    # Use folder entity to get inherited attribute values then.
+    if None in attr_values:
```
iLLiCiTiT:
```suggestion
    if None in attr_values.values():
```

MilaKudr (approved):
working well for me 

Image

